PR #267, opened 2015-02-17
CB-8500 Documentation update for signing property name change

nikhilkh:


agrieve:
Do we still want this change given that both gradle and ant-style are accepted? I'm thinking no since I find gradle-style makes more sense. If so, though, need to fix key.storeType -> key.store.type

nikhilkh:
I'm OK with not changing the docs here. I'll close this PR.

PR #285, opened 2015-05-07
CB-8752 Adds docs for hooks functionality

sgrebnov:
1. Moving hooks documentation to docs as this is general and not hello-world app specific
https://github.com/apache/cordova-app-hello-world/blob/master/hooks/README.md
2. Added Sample Usage section w/ end-to-end example


sgrebnov:
The following places must be updated to point to the docs after this is merged and released:
https://github.com/apache/cordova-app-hello-world/blob/master/hooks/README.md
https://github.com/apache/cordova-cli/#hooks-1

PR #292, opened 2015-05-22
CB-8900: Documentation for the platform save / restore feature

omefire:
This PR's goal is to document the platform save / restore feature.

### DIFF
--- a/docs/en/edge/platform_plugin_versioning_ref/index.md
+++ b/docs/en/edge/platform_plugin_versioning_ref/index.md
@@ -47,12 +47,12 @@ After running the above command, the resulting config.xml looks like :
 Some examples :
   * **'cordova platform add android --save'** => retrieves the pinned version of the android platform, adds it to the project and then updates config.xml.
   * **'cordova platform add android@3.7.0 --save'** => retrieves the android platform, version 3.7.0 from npm, adds it to the project and then updates config.xml.
-  * **'cordova platform add https://github.com/apache/cordova-android.git​ --save'** => clones the specified cordova-android git repository, adds the android platform to the project, then updates config.xml and point its version to the specified git-url.
+  * **'cordova platform add android@https://github.com/apache/cordova-android.git​ --save'** => clones the specified cordova-android git repository, adds the android platform to the project, then updates config.xml and point its version to the specified git-url.
   * **'cordova platform add C:/path/to/android/platform --save'** => retrieves the android platform from the specified directory, adds it to the project, then updates config.xml and point to the directory.
 
 ### Mass saving platforms on an existing project
-The '--save' flag describe above is only useful when you remember to use it during the platform addition.
-If you have a pre-existing project and you want to mass-save all existing platforms in it, you can use :
+The '--save' flag described above is only useful when you remember to use it during the platform addition.
+If you have a pre-existing project and you want to save all the currently added platforms in your project, you can use :
 
     $ cordova platform save
 
@@ -71,7 +71,7 @@ Some examples :
 
 ### Restoring platforms
   * Platforms are automatically restored from config.xml when the **'cordova prepare'** command is run.
-  * If you add a platform without specifying a version/folder/git_url, the version of the platform that would get added would be one compatible to the one in config.xml **if found**.
+  * If you add a platform without specifying a version/folder/git_url, the version to install is taken from config.xml, **if found**.
     * Example:
       Suppose your config.xml file contains the following entry:
           <?xml version='1.0' encoding='utf-8'?>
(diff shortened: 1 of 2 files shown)


omefire:
I'll soon send another one for the plugin side of things.

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,70 @@
...
+         "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+         KIND, either express or implied.  See the License for the
+         specific language governing permissions and limitations
+         under the License.
+---
+
+# Platforms and Plugins Version Management
+From version 4.3.0 onwards, Cordova provides the ability to save and restore platforms and plugins. <br />
+This feature allows developers to save and restore their app to a known state without having to check in all of the platform and plugin source code. <br />
+When a save command is issued, the config.xml file is used to reference details about platforms and plugins versions that should be used with the app. The 'restore' step happens automatically when a **'cordova prepare'** is issued. Restoration makes use of information previously saved in the config.xml file.
```
TimBarham: I don't think this reads well (`config.xml file is used to reference details`). Perhaps rearrange the sentence a bit - something like *The save command will store details about the app's platform and plugin versions in config.xml*. Also... not sure what our approach in docs is typically... but should command names be in monospace? As in `save` and `cordova prepare` etc?

TimBarham:
You should also describe what happens if you add a platform (via `cordova platform add`) that has an entry saved in `config.xml` (specifically that if you add the platform using some "@" spec, that will be used and will overwrite the entry in `config.xml`, and if you add the platform without any "@" spec, we will use whatever is saved in `config.xml`). Also, as of 5.0.0, a platform version is saved to `config.xml` using the caret prefix, which means you won't necessarily get the exact version you saved, but it should be a *compatible* version.

docs/en/edge/platform_plugin_versioning_ref/index.md
TimBarham: Instead of using a `<br>` here, can you just leave a blank line? The resulting paragraph spacing is much nicer. Like this:

From version 4.3.0 onwards, Cordova provides the ability to save and restore platforms and plugins.

This feature allows developers to save and restore their app to a known state without having to check in all of the platform and plugin source code.

purplecabbage: +1 <br/> has no place in a markdown file

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,70 @@
...
+The '--save' flag describe above is only useful when you remember to use it during the platform addition.
+If you have a pre-existing project and you want to mass-save all existing platforms in it, you can use :
+
+    $ cordova platform save
+
+
+### Updating / Removing platforms
+It is also possible to update/delete from config.xml during the commands 'cordova platform update' and 'cordova platform remove' :
+
+    $ cordova platform update <platform@<version>> --save
```
TimBarham: Strictly speaking, shouldn't this be:

    $ cordova platform update <platform[@<version>]> --save

(should it also include `directory` and `git_url` - does update work with those?)

omefire: Yes, it does.

omefire:
PR updated with code review comments !
Please, take another look.

omefire:
@TimBarham

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,82 @@
...
+        ...
+        <engine name="android" spec="^4.0.0" />
+        ...
+    </ xml>
+
+
+Some examples :
+  * **'cordova platform add android --save'** => retrieves the pinned version of the android platform, adds it to the project and then updates config.xml.
+  * **'cordova platform add android@3.7.0 --save'** => retrieves the android platform, version 3.7.0 from npm, adds it to the project and then updates config.xml.
+  * **'cordova platform add https://github.com/apache/cordova-android.git​ --save'** => clones the specified cordova-android git repository, adds the android platform to the project, then updates config.xml and point its version to the specified git-url.
```
TimBarham: Note that this doesn't actually work. When adding a platform from a location (other than a local folder), you currently have to specify the platform name. So this should be:

**cordova platform add android@https://github.com/apache/cordova-android.git​ --save**

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,82 @@
...
+  * **'cordova platform add C:/path/to/android/platform --save'** => retrieves the android platform from the specified directory, adds it to the project, then updates config.xml and point to the directory.
+
+### Mass saving platforms on an existing project
+The '--save' flag describe above is only useful when you remember to use it during the platform addition.
```
TimBarham: Typo: 'describe**d**'

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,82 @@
...
+If you have a pre-existing project and you want to mass-save all existing platforms in it, you can use :
```
TimBarham: "mass-save all" is a bit superfluous - how about just "you want to save all existing platforms" (although "existing" is perhaps a little unclear - perhaps "installed" or "currently added" or something).

docs/en/edge/platform_plugin_versioning_ref/index.md
```diff
@@ -0,0 +1,82 @@
...
+Some examples :
+  * **'cordova platform update android --save'** => In addition to updating the android platform to the pinned version, update config.xml entry
+  * **'cordova platform update android@3.8.0 --save'** => In addition to updating the android platform to version 3.8.0, update config.xml entry
+  * **'cordova platform update /path/to/android/platform --save'** => In addition to updating the android platform to version in the folder, update config.xml entry
+  * **'cordova platform remove android --save'** => Removes the android platform from the project and deletes its entry from config.xml.
+
+
+### Restoring platforms
+  * Platforms are automatically restored from config.xml when the **'cordova prepare'** command is run.
+  * If you add a platform without specifying a version/folder/git_url, the version of the platform that would get added would be one compatible to the one in config.xml **if found**.
```
TimBarham: How about something like:

If you add a platform without specifying a version/folder/git_url, the version to install is taken from config.xml, if found.

omefire:
Thanks @TimBarham  for reviewing. I've addressed your comments.

omefire:
I've also added documentation for plugins.

PR #295, opened 2015-07-07
CB-9295 Add GA analytics code on the Cordova Docs page

kant2002:
GA code added to all templates

### DIFF
--- a/template/docs/default/index.js
+++ b/template/docs/default/index.js
@@ -20,17 +20,13 @@
 //
 // Google Analytics
 //
-var _gaq = _gaq || [];
-_gaq.push(['_setAccount', 'UA-94271-11']);
-_gaq.push(['_setDomainName', '.phonegap.com']);
-_gaq.push(['_trackPageview']);
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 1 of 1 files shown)


nikhilkh:
Looks like https://github.com/apache/cordova-docs/blob/master/template/docs/default/index.js already has GA code. We should just re-use it and add the correct the ID.

nikhilkh:
Any ideas on how we can update older versions of the docs to use this new ID? Is it just a find and replace job?

kant2002:
If I will use code in the index.js, we have to update only index.js files across all versions and we are all set. (not sure about very old (<2.0) English versions. I will move code from index.html to index.js. This should be enough.

kant2002:
I also change from the old GA format to new. But we have to verify that both codes working. and as around about PhoneGap tracking code, do Adobe still use it?

One more thing which I could implement is add recording of events on following occasions
1. User change language
2. User select topic in the subheader select.

nikhilkh:
We don't care about the old GA tracking ID - in fact we do not have a way to even access it. As for the custom events, I'm new to google analytics but since they lead to navigations which are also tracked in GA, not sure if there is much value in adding custom events.

kant2002:
So if I could safely remove that old ID, I will do that. Any chance that somebody from Adobe has access and use it?

nikhilkh:
@stevengill tried to find if anyone could get access to it, but did not find the correct owner.

kant2002:
Ok, so I drop it.

kant2002:
To update the file index.js should be distributed to all versions.

nikhilkh:
I see - I can take a stab at updating all the older versions on the svn repo that has our website.

nikhilkh:
I merged this one 2f620e6. I had a typo in my commit message and this PR remained open. Can you please close it?

nikhilkh:
Thanks for the contribution!

PR #299, opened 2015-07-23
Corrected preference name in example

vikash:


dblotsky:
@vikash thanks for the PR! Sorry that it took us so long to get to it. There have been some major changes to the docs since this PR, and this error may have been fixed. Please take a look at [the new config.xml reference][reference] and let us know if the error is still there.

Thank you!

[reference]: http://cordova.apache.org/docs/en/latest/config_ref/index.html

PR #300, opened 2015-07-24
CB-9408: Documented the new `windows-packageVersion` attribute on

robpaveza:
... `<widget>` to allow Windows to have an independent binary version than its
widget.


omefire:
LGTM!

PR #306, opened 2015-08-19
Added plugin search js from cordova-registry-web

riknoll:
Adds the React javascript for the NPM plugin search from cordova-registry-web so that it can be integrated into the plugins page. The React components still require their HTML, CSS, and javascript to be updated before they can be used, but Gulp does already build app.js


dblotsky:
LGTM. Merged.

PR #312, opened 2015-08-25
Platform icons in search results and minor bug fixes

riknoll:
Added platform icons to plugin search results that show what platforms they support (thanks to @bgashler1) and made some minor fixes including:

* Search results now update when text is entered into the search bar without typing (e.g. right click and paste)
* Sort by download now handles the few platforms we don't have download counts for
* Download options/information do not appear until the counts have begun to come in


dblotsky:
Merged. For some reason GitHub didn't pick it up.

PR #323, opened 2015-08-29
Adding linting, link verifier and slack and twitter badges

rakatyal:


www/_includes/footer.html
```diff
@@ -72,3 +74,6 @@
 <script type="text/javascript" src="{{ site.baseurl }}/static/js/docs.js"></script>
 <script type="text/javascript" src="{{ site.baseurl }}/static/js/main.js"></script>
 <script type="text/javascript" src="{{ site.baseurl }}/static/js/index.js"></script>
+<script>
+    !function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0],p=/^http:/.test(d.location)?'http':'https';if(!d.getElementById(id)){js=d.createElement(s);js.id=id;js.src=p+'://platform.twitter.com/widgets.js';fjs.parentNode.insertBefore(js,fjs);}}(document, 'script', 'twitter-wjs');
```
dblotsky: What's this guy?

gulpfile.js
```diff
@@ -31,6 +33,7 @@ var WATCH_INTERVAL    = 1000; // in milliseconds
 var ROOT_DIR   = ".";
 var SOURCE_DIR = path.join(ROOT_DIR, "www");
 var DEV_DIR    = path.join(ROOT_DIR, "build-dev");
+var LINT_DIR    = path.join(DEV_DIR, "blog");
```
dblotsky: What is this directory?

gulpfile.js
```diff
@@ -247,6 +250,23 @@ gulp.task("link-bugs", function (done) {
     exec(bin("linkify-bugs.sh"), [path.join(SOURCE_DIR, "_posts")], done);
 });
 
+gulp.task('lint', function() {
+    return gulp.src(path.join(LINT_DIR, "**", "*.html"))
+        .pipe(htmllint());
+});
+
+gulp.task('checklinks', function(cb) {
+  crawler.crawl('http://localhost:3000/')
```
nikhilkh: This gulp task should depend on `serve` task. 
Nit: Also, the URL should be set for browser-link and used here as a constant - or get the default from browser-link if that's exposed instead of hardcoding this here.

gulpfile.js
```diff
@@ -247,6 +250,23 @@ gulp.task("link-bugs", function (done) {
...
+    .on('fetch404', function(queueItem, response) {
```
nikhilkh: Also, in our current website - almost all the broken links point to `#` which is technically valid but redundant- wonder if this is really caught by this tool - since it's not really a 404.

gulpfile.js
```diff
@@ -247,6 +250,23 @@ gulp.task("link-bugs", function (done) {
     exec(bin("linkify-bugs.sh"), [path.join(SOURCE_DIR, "_posts")], done);
 });
 
+gulp.task('lint', function() {
```
nikhilkh: Ideally, build should use this task and not succeed if linting fails. How about `jslint`?

www/_includes/footer.html
rakatyal: This is for the twitter icon added on the footer to follow Apache Cordova.

gulpfile.js
rakatyal: The directory where we want to run the lint. We can change it later as per use but currently the docs have a lot of formatting errors so I excluded it from linting.

dblotsky: But this excludes the rest of the site too, which should be linted. Is there a way to tell `htmlverify` to exclude/ignore a path?

gulpfile.js
dblotsky: Why should it depend on the `serve` task?

gulpfile.js
dblotsky: It's not caught. This tool can't replace manual inspection entirely.

gulpfile.js
dblotsky: Only when all linter errors are fixed, can we have build depend on lint. For now, it should be an independent task. We don't have jslint. I filed it as a separate feature to add, but it won't be in this PR.

gulpfile.js
nikhilkh: Because it expects localhost:3000 to be present for this to work

www/_includes/footer.html
dblotsky: Cool, thanks!

gulpfile.js
dblotsky: A server should be running for this task to succeed, but a server should not be started every time this task runs. Let's instead encode this dependency by adding a statement like `console.log("Make sure the server is running on " + SERVE_URL + "before running this task.")` to the `'checklinks'` task.

gulpfile.js
rakatyal: No. We would have to run the task for individual sub directories.

dblotsky:
Merged!

PR #346, opened 2015-09-15
Fixing consistency and grammatical errors on front page

riknoll:
Other possible improvements:

* In the first bullet, is "Terminal" capitalized for a reason?
* In the second bullet, "Create a Cordova project using the command-line tool to create a blank project" is redundant
* In the fourth bullet, "You can replace browser with the platform that you would like to run (e.g. ios, android, windows)" seems unnecessary (and if it is necessary, should also be mentioned in the preceding bullet)

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -93,8 +93,8 @@ permalink: /
         </div>
         <div class="col-md-7">
             <h2>Create a project</h2>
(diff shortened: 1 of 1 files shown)


dblotsky:
My thoughts:
- I think we're referring to Terminal.app on OS X, which is capitalized.
- Yep, the second "to create ..." is redundant and I'll remove it.
- I think we can remove that sentence; the previous point already gives the examples.

PR #354, opened 2015-09-18
Add Keyword Index back to the table of contents

riknoll:
Adds a translated keyword index page for each version of the docs that lists all the pages for that language/version. Page titles are scraped using toc.js

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -152,7 +152,7 @@ gulp.task("serve", ["build"], function () {
     });
 });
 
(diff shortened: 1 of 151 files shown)


riknoll:
@dblotsky can you review docs-keyword-index.js to make sure there aren't any security issues with how the page is taken from the Jekyll generated JSON and built?

riknoll:
and the docs_index.html _include as well

dblotsky:
Yep, LGTM. There is no user input, so there shouldn't be any security issues.

dblotsky:
Feel free to close.

PR #355, opened 2015-09-18
Minor layout fixes for home and docs

riknoll:
Adds padding to the not latest version text in docs

Also fixes the "+4 more..." text on the front page so that it doesn't wrap awkwardly on small screens(e.g. "+4" on one line and "more..." spilling over to the next).


dblotsky:
Looks like I merged this, and again forgot to mention that. :/ Sorry, my bad.

PR #356, opened 2015-09-21
Add browser api fallback for copy buttons

riknoll:
ZeroClipboard is used on the main and plugins pages for the copy buttons and relies on Flash to operate. Modern browsers are gaining support for executing copy commands, so we can use that API as a fallback when Flash is disabled. In the future we should also add something to account for both failing. However, this works on Chrome, IE11, and Edge at least.


riknoll:
It is apparently not supported by Safari or current versions of Firefox (though it will be added in 41.0 when it comes tomorrow). https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

dblotsky:
Merged. Feel free to close.

PR #357, opened 2015-09-21
Made the first heading in each docs page the page title

riknoll:
Accomplished using the same regex that generates the keyword index page. I did not personally check all 9,125 docs files that changed, so merge with caution. Also adds the ability to give a title prefix to a page (these need to be translated for all of the languages in the docs)


dblotsky:
Merged. Feel free to close!

PR #358, opened 2015-09-23
Implementing Proper Latest Docs Version

dblotsky:
- Adding docs root and language root redirects.
- Adding `site.latest_version` for that purpose, and for `not_latest_warning_text`.
- Changing all links to `docs/{lang}/edge` to `docs/{lang}/{latest_version}` because that is what they actually should be.


riknoll:
Looks good to me!

PR #359, opened 2015-09-23
Adding Syntax Highlighting to Site

dblotsky:


riknoll:
Looks good to me!

_config.yml
```diff
@@ -26,7 +26,13 @@ keep_files: [".git", ".svn", "wiki-images", "images", "downloads"]
 lsi:  false # don't produce an index for related posts
 safe: true  # disables plugins
 
-markdown: kramdown
+markdown: redcarpet
+
+redcarpet:
+    extensions:
+        - prettify
+        - fenced_code_blocks
```
riknoll: Nitpick: fenced_code_blocks is set to be true by default in Jekyll. We don't need to specify it: http://jekyllrb.com/docs/configuration/#redcarpet

dblotsky: If that default behaviour changes, specifying it here makes it clear that we explicitly want it. I'll add a comment to explain this.

PR #362, opened 2015-09-28
Improving In-Development ("Edge") Page Logic

dblotsky:
- Added a "this is edge" message to edge versions.
- Made nonexistent docs versions still clickable.
- Now there are two messages: "this is edge", and "this is not the latest".

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,7 +12,7 @@ repo:
     uri:    https://github.com/apache/cordova-docs
     branch: cordova-website
 
-latest_version: 5.1.1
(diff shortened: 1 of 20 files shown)


riknoll:
This looks fine to me. My only comment is that the in development text does not make it immediately clear that the warning bubble is a link or what clicking on it will do. It might be clearer if we added "The current version is 5.1.1" or something to that effect to the end of it.

riknoll:
Looks good to me.

dblotsky:
Thanks for the review!

PR #363, opened 2015-09-28
Improving 404 for Missing Versions

dblotsky:
- Now displaying a message for missing docs versions, linking to "edge."
- Still showing "moved here" message for 404s on existing versions.


riknoll:
Interestingly, this is broken in Edge (the browser, not the version). I cannot get the not-translated warning to appear on paths like /docs/ko/5.1.1/index.html (the moved here redirect does appear correctly). Works in IE and Chrome. Must be an AJAX problem.

dblotsky:
Thanks for the review! I'll merge this and make a separate task for the Edge bug.

PR #395, opened 2015-10-19
New Cordova Website and Docs - Rewritten

dblotsky:
Merging `cordova-website` branch by copying over files and committing them anew.


brodybits:
Any pointers where we could find the `cordova-website` branch?

It would really help us to have the history and attribution of the files that were added in this PR, if possible.

dblotsky:
I’m not sure it exists anymore. Cordova docs and website used to be separate repos, and became one.

I might have some context though. What’s up?

Dmitry

brodybits:
I was wondering if there may be any pointers where we could find an archive of the old website repo?

Shouldn't the ASF git keep this kind of an archive somewhere?

I think this could help us understand what was done in the past, why, and by whom. Discussion of the flash artifact in #1031 is a prime example. I think maintaining contributor credits whenever possible is part of the general ASF philosophy.

Thanks for your response here and in #1031.

dblotsky:
I’ll dig a bit when I’m at home today and let you know what I find!

Dmitry

brodybits:
Here is how we can look through the history of cordova-website, starting from the beginning (see <https://github.com/apache/cordova-docs/issues/1033#issuecomment-565142797>):

```sh
svn checkout https://svn.apache.org/repos/asf/cordova/site cordova-website
cd cordova-website
svn log -r 1:HEAD -v | more
```

PR #397, opened 2015-10-20
Small typo on line 100.

KSCooke:
Line 100 read "The plugin commands are a mirror of the plugin commands" and I think the second plugin commands was intended to be platform commands. Either way it's true.


dblotsky:
Merged. Thanks for the PR!

PR #402, opened 2015-10-23
Remove misleading reference to "publisherId" cause it is ignored

pke:
The only way to override the `Publisher` value in the generated `AppxManifest.xml` is to provide a custom signing certificate. Merely setting the "publisherId" does not do anything. I've tried it a whole day. Only when I set a custom signing certificate my applications manifest looked like it should. And then setting "publisherId" to anything did not affect how the resulting manifest file looked.

So I propose to remove the value from the docs (and it should also be removed from the Cordova Windows plugin).

### DIFF
--- a/www/docs/en/5.1.1/guide/platforms/win8/packaging.md
+++ b/www/docs/en/5.1.1/guide/platforms/win8/packaging.md
@@ -35,7 +35,7 @@ In Windows project, identity details are kept in a file named package.appxmanife
 - Publisher
 - Version
 
-*Name* and *Version* can be set from **config.xml**. *Publisher* can be provided as a build parameter or can be set on **build.json** file.
+*Name* and *Version* can be set from **config.xml**. *Publisher* can be provided by a signing certificate `.pfx` as a build parameter or can be set on **build.json** file.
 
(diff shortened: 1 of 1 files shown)


dblotsky:
@stevengill or @purplecabbage: I'm not very familiar with this topic. Could either one of you please review this when you have a moment? Or point to someone that has knowledge in this area? Thanks, guys.

stevengill:
I think @nikhilkh might have more info on this

dblotsky:
@nikhilkh ping; please give this a look when you have a moment.

purplecabbage:
Is this still an issue?  Did you find a suitable solution @pke ?

pke:
This PR has nothing to do with Windows but mentions Android

purplecabbage:
One commit, a merge and a close ... unrelated.
Much like 25485819612937251bb521a9b68f73cae54065c3

pke:
Whats going on here? Is @asfgit a failing bot? My PR is still valid

purplecabbage:
No, I closed it. Sorry, I didn't think it was still valid.

PR #405, opened 2015-10-28
Adding support for /docs/XX/latest

dblotsky:
Adding support for `latest` URI, which always points to the current latest docs version.


dblotsky:
Food for thought: is it better just to 302-redirect to the version?

PR #413, opened 2015-11-05
Switch to /latest

dblotsky:
Switching to using /latest URI by default. Fixing 404 page to recognize it.


dblotsky:
@riknoll please review

PR #414, opened 2015-11-05
Set proper name in iOS config documentation

asc0de:
Correct property name in example is wrong


dblotsky:
Thanks for the PR @asc0de! I merged it in and followed up with a change to all the other languages.

PR #415, opened 2015-11-06
Default value for required variables

mbektchiev:


stevengill:
Just confirming that plugman already handles the default attribute on preference tags before I merge this in.

mbektchiev:
Here's the JIRA issue which added this support: https://issues.apache.org/jira/browse/CB-9162

stevengill:
Awesome. Thanks for the commit!

PR #422, opened 2015-11-16
CB-10017: Restored broken image in amazon fireos docs

riknoll:
This restores a broken image link that was deleted as part of f90b9d66056349aa33480675f013ae9269823d8d. Clearly, this is a big problem with how we handle images. Images need to be pinned with versions, otherwise if an image is deleted in master then the link will be broken in all of the past docs (hopefully we can come up with a smart way of handling this that does not require us to have 50 copies of every image). We should probably do a sweep to see if there are any other missing images in the past docs. Also, the linked commit makes it seem like this image isn't currently correct for Amazon fireos.


riknoll:
Maybe we could include a test script to find broken images as part of the deployment/release process?

riknoll:
@dblotsky please review

dblotsky:
LGTM. I'll create a JIRA for the test script so that we can address it in the future.

PR #418, opened 2015-11-13
Added page describing how to create issues

riknoll:
Replaces the [old wiki page](http://wiki.apache.org/cordova/ReportingBugs) with a new page based mostly on the [JIRA Triage docs](https://github.com/apache/cordova-coho/blob/master/docs/jira-triage.md)

PR #423, opened 2015-11-17
Refactored the title prefix into a global prefix in _config.yml

riknoll:
No JIRA since this is just a refactor. Moves all the title prefixes into the global _config.yml for consistency's sake and fixes a few page titles as well (contribute was named plugins). Note that this also affects the titles in #418


riknoll:
@dblotsky please review

dblotsky:
One note: do you think it might be better to have `title_prefix` set to nothing on blog posts? Or maybe to change the prefix logic somehow to make titles visibly unique in the browser? On Chrome if I have a page open, I mostly see `Apache Cordova - ...` in the title, which seems off.

riknoll:
I noticed the same thing. I don't know why we have one, but I remember us making the decision at some point... Does it have to do with SEO or something? Should we put it at the end of the title?

dblotsky:
Putting it at the end sounds like a good idea to me. I guess then it would just be more appropriate to call it `title_suffix`.

riknoll:
Okay, updating PR

PR #424, opened 2015-11-17
CB-10020: Indicated the latest version in the docs version dropdown

riknoll:
Adds "(Latest)" next to the latest version in the version dropdown


riknoll:
@dblotsky please review

dblotsky:
Upon shrinking the screen size, I noticed that the `(Latest)` string makes the dropdown go outside the bounds of the screen. I was thinking of doing something like making the latest version just appear different in the dropdown (bold, underlined, coloured). What do you think?

riknoll:
I don't think it is immediately clear what that would mean. We already highlight the current version, but now there are two highlighted versions? Which is which? What does highlighting even mean? I thought just labelling it would be clearer. Could we maybe drop the version altogether and just put latest?

riknoll:
Also, I think the [live site](http://cordova.apache.org/docs/en/latest/guide/overview/) has the same dropdown issue. The `min-width` is actually being set for the `.drop-down` class, which is what is causing the issue (not the `latest` string). I can update my PR to remove the `min-width`, but the dropdown still goes outside the bounds sometimes.

riknoll:
@dblotsky updated to fix the dropdown issue. Note: This also affects the language dropdown

PR #427, opened 2015-11-19
CB-10045: Made table elements in docs extend bootstrap table classes

riknoll:
I'm adding a docs page and need table styles. This affects the platform support table FYI. @dblotsky please review.


dblotsky:
LGTM.

PR #429, opened 2015-11-21
Website fixes

dblotsky:
Each commit is a separate improvement. In general, this is what changed:
- version dropdown now says "(Latest)" in a translated manner
- some pages now have weights and update frequencies for `sitemap.xml`
- there is now a redirect from `edge/` to `dev/` in `.htaccess`
- the font on the header of the table on [this page](http://cordova.apache.org/docs/en/dev/guide/support/index.html) is no longer monospace
- `contribute.html`, `blog.html`, and `plugins.html` layouts now inherit from the `page.html` layout


dblotsky:
@riknoll please review when you have a moment!

PR #430, opened 2015-11-21
Fixing some styling issues and simplifying content on Contribute page.

dblotsky:
Went from this:
old
to this:
new.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,20 +194,34 @@ $(TOC_FILES): $(BIN_DIR)/toc.js Makefile
 
 $(MAIN_STYLE_FILE): $(SCSS_SRC)
 
+# NODE:
+#      $(@D) means "directory part of target"
 $(CSS_DEST_DIR)/%.css: $(CSS_SRC_DIR)/%.less Makefile
+ifdef WINDOWS
 	-$(MKDIRP) $(subst /,\,$(@D))
+else
+	$(MKDIRP) $(@D)
+endif
 	echo ---> $@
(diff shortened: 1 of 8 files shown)


purplecabbage:
Looks great!
I think we should also drop the Apache-Git links completely.  All bug fixes, and pull requests can be handled completely without ever knowing that we are using github mirrors ... and I suspect it might confuse newcomers when they are dumped unmercifully into an apache-git page.

purplecabbage:
nit: Can we also change the 'non-committers' to 'contributors' or something, so they still sound like they have commitment?

dblotsky:
To be honest, it's been helpful even for me to get the ASF Git links by going to that page. I'd personally prefer to keep them there, but I don't feel too strongly about it. Unless you or others feel strongly about removing them, I'd suggest just keeping them there. As for the link, how about "Committers / Other Contributors"?

stevengill:
+1 to changing non committers to contributors. That is how apache classifies the difference.

I am fine leaving the apache git urls. Most people will click github icon anyways.

How do you feel about adding a join us on slack on that page?

purplecabbage:
Hint: all repos are listed here, my fave bookmark. 
https://git-wip-us.apache.org/repos/asf?a=project_list&s=cordova-&btnS=Search

When you follow a github link, the first thing you see is a nicely formatted readme which is a great introduction. 

I think we should put the slack link *everywhere*, including the top of every github repo readme.md

PR #435, opened 2015-12-04
Adding Search Button

dblotsky:
Uncommenting search icon and linking it to Google search.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,6 +194,8 @@ $(TOC_FILES): $(BIN_DIR)/toc.js Makefile
 
 $(MAIN_STYLE_FILE): $(SCSS_SRC)
 
+# pattern rules
+
 # NODE:
 #      $(@D) means "directory part of target"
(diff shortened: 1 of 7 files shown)


dblotsky:
@riknoll please review when you have a moment.

riknoll:
@dblotsky That search icon is barely visible in the top right; I don't think people will notice it. Can we add a text input to make it easier to see and use? The icon also looks a bit awkward in the mobile nav menu. It adds itself to the expandable list and is the only non-text option there. I think that for a mobile screen, the search icon should stay in the top bar and appear to the left of the menu-expand icon (no need for a text input for mobile, it would just clutter the nav bar). Thoughts?

dblotsky:
@riknoll I've incorporated your feedback. Please give it a look when you have a moment.

dblotsky:
@riknoll and I've again updated the PR. Please give it a look when you have a chance. In particular, on `md` screens the search field disappears, and the search button now sticks to the right of the field.

riknoll:
LGTM

PR #436, opened 2015-12-08
Add cordova-ios-4.0.0 release blog post announcement

shazron:


dpogue:
Might be good to include a line like "This is a major release, and deprecated APIs have been removed. Some 3rd party plugins might require updates before they are compatible."

shazron:
@dpogue  Good idea, will add that in.

csantanapr:
What about mentioning other some high level features:

1. Support Asset Catalog
2. ios-sim (will use internal version, no need to npm i -g ios-sim)
3. ios-deploy (get the latest version)
4. can handle new fields in config.xml (I'm making this up? like xxconfig additions)

shazron:
@csantanapr good points, I'll add those in. Not sure about point 4 though, xcconfig stuff is build artifacts, and we don't have the `<build-setting>` tag yet for this

csantanapr:
@shazron ah ok, maybe I was thinking <build-setting>

oh what about whitelist stuff?

Any other high level feature?

csantanapr:
What about AppDelegate changes/moving?

What about minimum Deployment Target to iOS 8.0? What does this means for humans :-), no ios 7?

shazron:
@csantanapr whitelist stuff is already mentioned.

Deployment target change -- yes. Too many other changes to list in one blog post, that's why there is a "breaking changes" link that lists all.

csantanapr:
ok, I went over the wholes changes, that's it I didn't find anything else

shazron:
Ok all changes integrated, for those latecomers here is the link to view all PR changes: https://github.com/shazron/cordova-docs/blob/cordova-ios-4-release/www/_posts/2015-12-08-cordova-ios-4.0.0.md

csantanapr:
LGTM

PR #438, opened 2015-12-10
'ccordova' is a typo

fanfare:
changed ccordova to cordova


dblotsky:
@fanfare sorry for the delay in responding to you! Thank you very much for your pull request. This change had already been applied between the time you opened the PR and we got to merging it. However I've taken your change and applied it across all docs, so that now there are no more mentions of `ccordova` anywhere. :) Thank you again for contributing!

PR #459, opened 2016-01-19
Fix invalid dash sign (`Too many parameters` exception)

SunboX:


dblotsky:
Thank you, @SunboX for your contribution! Sorry that it took us so long to merge your change.

PR #460, opened 2016-01-21
CB-10330 Adding style guide for xml references

rakatyal:


nikhilkh:
This looks great! Couple of comments - while your example illustrates this - your specification is not very precise:
1. Consider clarifying the position of 'Required', 'Default' in the description. Also, consider structure the details e.g. Description should include the following in order:
- Default: Use *italics* formatting to provide default value of the attribute
- Required: Use *italics* formatting to provide whether the attribute is required. I would imagine if a value is not required - there would be a default value.
- Platforms supported: Use **bold** to specify platforms supported. Default is supported in all platforms.

2. Consider making the platform supported part of the description instead of the name. You can still use bold.

riknoll:
The content looks good to me, but I had difficulty reading the document. It wasn't clear to me which parts of it are examples and which are the guidelines. The examples are all inline and they include headers, which makes everything a bit confusing because the example appears bigger than the guideline it corresponds to. It would be nice if they could be differentiated in some way (maybe the examples could be put in a "well" like a code block is).

omefire:
Looks good, but I'm a bit confused with examples, which ones are examples ? which aren't ?

rakatyal:
Addressed comments.

nikhilkh:
LGTM

riknoll:
LGTM

PR #461, opened 2016-01-21
CB-8917: Improving Android lifecycle docs

riknoll:
@nikhilkh @jasongin please give this a review when you get the chance. I rewrote the lifecycle guide to clarify a few things. Mostly, I wanted to emphasize what situations the lifecycle is relevant in and what the lifecycle actually refers to.

### DIFF
--- a/www/docs/en/dev/guide/platforms/android/lifecycle.md
+++ b/www/docs/en/dev/guide/platforms/android/lifecycle.md
@@ -22,57 +22,127 @@ title: Android Lifecycle Guide
 
 # Android Lifecycle Guide
 
-Native Android apps typically consist of a series of "activities" that the user interacts with. Activities can be thought of as the individual screens that make up an application; different tasks in an app will often have their own activity. Each activity has a lifecycle, and can be created and destroyed as the OS sees fit.
+## Cordova and Android
 
-In contrast, Cordova applications on the Android platform are executed within a Webview that is embedded in a *single* Android activity. The lifecycle of this activity is exposed to your application through the document events that are fired. These events are not guaranteed to line up with Android's lifecycle, but they can provide guidelines for saving and restoring your state. These events roughly map to Android callbacks as follows:
+Native Android apps typically consist of a series of [activities](http://developer.android.com/reference/android/app/Activity.html) that the user
+interacts with. Activities can be thought of as the individual screens that make
+up an application; different tasks in an app will often have their own activity.
+Each activity has its own lifecycle that is maintained as the activity enters
+and leaves the foreground of a user's device.
 
-Cordova Event   | Rough Android Equivalent
-----------------|-----------------------------------
-`deviceready`   | `onCreate()`
-`pause`         | `onPause()`
-`resume`        | `onResume()`
+In contrast, Cordova applications on the Android platform are executed within a
+Webview that is embedded in a *single* Android activity. The lifecycle of this
+activity is exposed to your application through the document events that are
+fired. The events are not guaranteed to line up with Android's lifecycle, but
+they can provide guidelines for saving and restoring your state. These events
+roughly map to Android callbacks as follows:
 
-Often, Cordova applications are confined to the single activity that contains the Webview. However, there are instances in which other activities may be launched that push the Cordova activity to the background. It is important in these cases to be aware of the Android lifecycle and properly maintain your application's state by respecting it.
+Cordova Event   | Rough Android Equivalent  | Meaning
+----------------|---------------------------|-----------------
+`deviceready`   | `onCreate()`              | Application is starting (not from background)
+`pause`         | `onPause()`               | Application is moving to the background
+`resume`        | `onResume()`              | Application is returning to the foreground
 
-## Low memory and the Activity lifecycle
+Most other Cordova platforms have a similar concept of lifecycles and should
+fire these same events when similar actions happen on a user's device. However,
+Android presents some unique challenges that can sometimes show up thanks to the
+native Activity lifecycle.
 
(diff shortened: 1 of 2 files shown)


www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -80,14 +150,20 @@ The payload for the `resume` event adheres to the following format:
...
+        pluginStatus: string,
+        result: any
     }
 }
 ```
 
-The possible plugin statuses in the `pendingResult` field include the following values:
+The fields of that payload are defined as follows:
+
+* `pluginServiceName`: The name of the plugin return the result (e.g. "Camera"). This can be found in a plugin's plugin.xml
```
nikhilkh: Where in the plugin.xml do I find this name?

www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -102,12 +178,16 @@ The possible plugin statuses in the `pendingResult` field include the following
 "Error
 ```
 
->**NOTE:** It is up to the plugin to decide what is contained in the `result` field and the meaning of the `pluginStatus` that is returned. Reference the API of the plugin you are using to see what you should expect those fields to contain and how to use their values
+Please note that it is up to the plugin to decide what is contained in the
```
nikhilkh: Does the camera plugin have details on what these results mean and how one should respond?

www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -80,14 +150,20 @@ The payload for the `resume` event adheres to the following format:
...
 ```
 
-The possible plugin statuses in the `pendingResult` field include the following values:
+The fields of that payload are defined as follows:
+
+* `pluginServiceName`: The name of the plugin return the result (e.g. "Camera"). This can be found in a plugin's plugin.xml
+* `pluginStatus`: The status of the plugin call (see below)
+* `result`: Whatever the result of the plugin call is
+
+The possible values for `pluginStatus` in the `pendingResult` field include the following:
```
nikhilkh: Remind me again - why the pluginStatus is a string and not some sort of enum? It will help to explain what each of these statuses mean.

www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -22,57 +22,127 @@ title: Android Lifecycle Guide
...
+the result of that plugin call when your application resumes.
 
 ## Retrieving plugin callback results
 
-When the OS destroys the Cordova activity in the above example, any pending callbacks are lost as well. This means that if you passed a callback to the plugin that launched the new activity (e.g. cordova-plugin-camera), that callback will NOT be fired when the application is recreated. However, there is a way for plugins to pass the result of this call to your application. The `resume` event's payload will contain any pending plugin results from the plugin request that launched the external activity made prior to the activity being destroyed.
+When the OS destroys the Cordova activity that was pushed into the background
+by a plugin, any pending callbacks are lost as well. This means that if you
+passed a callback to the plugin that launched the new activity (e.g. the camera
+plugin), that callback will NOT be fired when the application is recreated.
+However, starting in cordova-android 5.1.0, the `resume` event's payload will
```
nikhilkh: The version needs to be in BOLD and distinctive - perhaps as a top-level heading

www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -102,12 +178,16 @@ The possible plugin statuses in the `pendingResult` field include the following
...
 ```
 
->**NOTE:** It is up to the plugin to decide what is contained in the `result` field and the meaning of the `pluginStatus` that is returned. Reference the API of the plugin you are using to see what you should expect those fields to contain and how to use their values
+Please note that it is up to the plugin to decide what is contained in the
+`result` field and the meaning of the `pluginStatus` that is returned. Reference
+the API of the plugin you are using to see what you should expect those fields
+to contain and how to use their values.
 
-A `resume` callback that makes use of the event payload might look something like this:
+A `resume` callback that makes use of a plugin's event payload might look
```
nikhilkh: Consider adding an Example heading.

www/docs/en/dev/guide/platforms/android/lifecycle.md
```diff
@@ -116,10 +196,21 @@ onResume(event) {
     if(event.pendingResult) {
         // Figure out whether or not the plugin call was successful
         if(event.pendingResult.pluginStatus === "OK") {
-            successCallback(event.result);
+            successCallback(event.pendingResult.result);
```
nikhilkh: What does this successCallback refer to? Perhaps you want to give a realistic example with cordova-plugin-camera. I would recommend including the same example in cordova-pluign-camera docs.

nikhilkh:
It will be useful to add references to Android docs.

www/docs/en/dev/guide/platforms/android/lifecycle.md
riknoll: There is no real reason it is a string vs an enum. I tried to clarify below that the meaning of statuses are for the most part determined by the author of the plugin. The other ones (class not found, etc.) are really intended for plugin developers and don't really make sense to expose to the user. However, we don't really have a choice in exposing them thanks to how cordova-android handles plugin results. I will try and define them here, but I think plugin developers should be the ones to specify what to check for.

nikhilkh: I guess there are two buckets - "OK" and everything else which implies an error - correct?

riknoll: And "No Result", but yeah that is the general idea. I can still add a short sentence for each

riknoll:
@nikhilkh I added a much more detailed example and incorporated your other feedback

PR #464, opened 2016-01-22
Multi-level ToC Implementation

dblotsky:
- Added support for nested ToCs using the `children` attribute
- Retroactively made past ToC files use relative links
- Made all dev ToCs use the new documentation information architecture

### DIFF
--- a/www/_data/toc/de-dev-manual.yml
+++ b/www/_data/toc/de-dev-manual.yml
@@ -1,19 +1,70 @@
-- {name: "Reisef\xFChrer", url: !!python/unicode '/docs/de/dev/'}
-- {name: "\xDCbersicht", url: !!python/unicode '/docs/de/dev/guide/overview/index.html'}
-- {name: "Plattformunterst\xFCtzung", url: !!python/unicode '/docs/de/dev/guide/support/index.html'}
-- {name: !!python/unicode 'Die Befehlszeilenschnittstelle', url: !!python/unicode '/docs/de/dev/guide/cli/index.html'}
-- {name: !!python/unicode 'Plattform-Guides', url: !!python/unicode '/docs/de/dev/guide/platforms/index.html'}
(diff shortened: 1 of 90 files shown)


www/_data/toc/de-dev-manual.yml
```diff
@@ -1,19 +1,70 @@
...
+        -   name: "Tooling Architecture"
+            url:
+-   name: "Create apps"
+    children:
+        -   name: "Create your first app"
+            url:
+        -   name: "Develop for platforms"
+            url: "guide/platforms/index.html"
+        -   name: "Manage versions and platforms"
+            url:
```
rakatyal: This article is present in current TOC. platform_plugin_versioning_ref/index.html

nikhilkh:
If an article does not have a URL do we make sure it does not show up in the displayed page?

nikhilkh:
Looks like there are really two changes - use of relative links & multi-level TOC. It's hard to follow but I'm assuming only newer versions of the docs do get multi-level TOC - older versions remain unchanged.

dblotsky:
@nikhilkh I made articles not show up if they don't have a URL. As for older versions, they were only updated to use relative links.

www/_data/toc/de-dev-manual.yml
dblotsky: Got it. Thanks!

nikhilkh:
This looks great! One issue: "Architecture" has no children with URLs and it shows up in the TOC. 
image

dblotsky:
@nikhilkh that is a difficult bug to fix because it relies on knowing that all children of "Architecture" also don't have any URIs. It will start to have some pages under it very soon, so I think it's ok to have the title be empty in the short term.

nikhilkh:
Could you comment "Architecture" and it's sub-sections? We might not get to write these topics till April or something. It's best to not to have empty non-clickable placeholders in TOC.

dblotsky:
@nikhilkh Fixed. Merging.

PR #466, opened 2016-01-26
CB-10440 Add CSS color names support for BackgroundColor on Windows

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10440)

### DIFF
--- a/www/docs/en/dev/config_ref/index.md
+++ b/www/docs/en/dev/config_ref/index.md
@@ -167,6 +167,15 @@ all of them:
 
         <preference name="BackgroundColor" value="0xff0000ff"/>
 
+  Windows Quirk: CSS Color names are also supported:
+
+  ```
+    <platform name="windows">
+        <preference name="BackgroundColor" value="turquoise"/>
+    </platform>
+  ```
+  __NOTE__: `transparent` value will cause black/white background color in case of Dark/Light theme accordingly.
+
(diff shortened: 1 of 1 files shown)


www/docs/en/dev/config_ref/index.md
```diff
@@ -167,6 +167,17 @@ all of them:
...
+- CSS Color names are also supported:
+
+```
+<platform name="windows">
+    <preference name="BackgroundColor" value="turquoise"/>
+</platform>
+```
+  __NOTE__: `transparent` value will cause black/white background color in case of Dark/Light theme accordingly.
+
   Applies to Android and BlackBerry. Overrides CSS otherwise available
```
nikhilkh: Does the 'applies to Android & Blackberry' needs to be fixed?

daserge: Yes, thanks, I've just confused this notation with the plugin docs one where we have a supported platforms list in the start and Quirks in the end.

daserge: @nikhilkh, Actually I believe the phrase `Applies to Android and BlackBerry. Overrides CSS otherwise available across _all_ platforms, for example: `body{background-color:blue}`.` says about page style taken from `BackgroundColor` preference, which is not supported on Windows.
I will update `Applies to Windows.` to `Windows Quirk` to avoid confusing with the rest of the doc though.

PR #472, opened 2016-01-29
Sitemap Improvements

dblotsky:
- Implementing sitemap.xml manually instead of using a plugin. 
- Adding /latest/ URIs to sitemap.xml. Moving some CSS from main.scss to _docs.scss.


dblotsky:
@riknoll @nikhilkh please review.

nikhilkh:
I'm curious what's the motivation for not using the plugin?

dblotsky:
@nikhilkh The plugin does not support adding custom URIs to the sitemap, and extending it took more effort than just writing a templated config.xml.

nikhilkh:
Fair enough - let's get it in!

PR #473, opened 2016-01-29
Changing the XML reference formatting for required and platforms attr…

rakatyal:
…ibutes

### DIFF
--- a/www/static/css-src/main.scss
+++ b/www/static/css-src/main.scss
@@ -173,6 +173,12 @@ code.prettyprint {
     border: none !important;
 }
 
+mark {
+    background-color: lightblue;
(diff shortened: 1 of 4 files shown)


www/static/css-src/main.scss
```diff
@@ -173,6 +173,12 @@ code.prettyprint {
     border: none !important;
 }
 
+mark {
+    background-color: lightblue;
+    padding: 0.1px 0.2px 0.1px 0.2px;
+    margin-right: 7px;
+}
+
```
riknoll: This should probably go in `_docs.scss`

rakatyal: Right.

nikhilkh:
LGTM

riknoll:
LGTM

PR #476, opened 2016-02-01
Do not escape markdown code for relative links

mikrobi:
The links were not clickable. The markdown code was just shown in plaintext.


dblotsky:
@mikrobi thank you for your PR! Sorry that it took us so long to get to merging it. We've also made note of this type of link in our source code, and we'll make a sweeping change to fix them all in the near future.

PR #479, opened 2016-02-02
CB-10503: Documentation link points to 'latest' version even in latest buils

sarangan12:
Fixed the link to point to dev version of the docs in development

@dblotsky @riknoll @rakatyal @nikhilkh Can you please review and merge this PR?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -22,6 +22,11 @@ repo:
 # NOTE:
 #      this value gets overwritten in _version.yml, which is a generated file
 latest_docs_version: dev
+# NOTE:
+#      this value is used to point to the documentation version to be used. 
+#      in development this may point to dev (in _dev.yml) and in prod this
+#      may point to latest (in _prod.yml).
+default_linked_docs_version: "latest"
 
 # NOTE:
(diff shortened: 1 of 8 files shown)


_config.yml
```diff
@@ -9,6 +9,7 @@ title:          Apache Cordova
 title_suffix:   "Apache Cordova"
 url:            "https://cordova.apache.org" # the base hostname & protocol
 baseurl:        "" # the subpath of the site, e.g. /blog; NOTE: no trailing slash
+docsversion:       "latest"
```
riknoll: Whitespace is weird here. Otherwise LGTM

sarangan12:
@riknoll Fixed the spacing

dblotsky:
We have another variable, `latest_docs_version`, whose name sounds like this one. Please rename this variable to something more explicit (like `default_linked_docs_version`) and please add it above [this line][line] so that it's grouped with the other one.

[line]: https://github.com/apache/cordova-docs/blob/master/_config.yml#L22

riknoll:
+1 to @dblotsky's comment

PR #480, opened 2016-02-02
CB-10351: Make Events page a reference

sarangan12:
Made the Events page a reference

@nikhilkh @dblotsky @rakatyal @riknoll @omefire Can you please review and merge this PR?


riknoll:
Resume event should add an Android quirk that points to this page: http://cordova.apache.org/docs/en/dev/guide/platforms/android/lifecycle.html

sarangan12:
@riknoll Added link to Android Life Cycle Guide

www/docs/en/dev/cordova/events/events.md
```diff
@@ -22,29 +22,561 @@ title: Events
...
+
+    // device APIs are available
+    //
+    function onDeviceReady() {
+        // Now safe to use device APIs
+    }
+
+## pause
+
+The pause event fires when an application is put into the background. The `pause`
```
rakatyal: Line 1 and 2 are repetitive.

sarangan12: That was intentional. To explicitly say, this is just a JS file and also what the following function is used.

rakatyal: I meant the description of 'pause' event. We can probably make it a single line.

sarangan12: Oops. Fixed it now

rakatyal:
LGTM.

www/docs/en/dev/cordova/events/events.md
```diff
@@ -22,29 +22,560 @@ title: Events
...
+        document.addEventListener("resume", onResume, false);
+        function onResume() {
+           setTimeout(function() {
+                  // TODO: do your thing!
+                }, 0);
+        }
+
+### Android Quirks
+
+Refer [Android Life Cycle Guide][AndroidLifeCycleGuide] for details on android quirks for
```
riknoll: Minor formatting/grammar edit:

Refer to the [Android Life Cycle Guide][AndroidLifeCycleGuide] for details on Android quirks with the `resume` event.

riknoll:
I mentioned this offline, but do we want to include the full example for each event? Adds a lot of length to the document and all of the examples are the same with strings swapped. I think the quick examples are sufficient.

www/static/js/docs.js
```diff
@@ -40,7 +40,7 @@ $(document).ready(function () {
 
     // Table of Contents
     $('#page-toc').toc({
-        'selectors':         'h1,h2,h3,h4,h5,h6', // elements to use as headings
+        'selectors':         'h1,h2', // elements to use as headings
```
dblotsky: Why not have `h3,h4,h5,h6` show up in the page ToC?

dblotsky:
+1 to Richard's previous comment.

rakatyal:
+1 to Richard's point. They hardly add anything relevant and the format repeats for each event. I think that we can rename it to example, give the full example for the first event and keep the quick example content for others to follow.

www/static/js/docs.js
riknoll: I didn't notice this. We do need this change, but it should be in a separate commit/JIRA. The reason for this @dblotsky is that certain longer pages (like the upcoming single page platforms guide) have a lot of headers that only serve to clutter the toc. Same with reference pages; every attribute does not need a toc entry.

dblotsky: I guess that's an ok workaround. But doesn't that mean that we should use different tags other than headers in the references? This might become confusing when we do use `h3`s properly in a document and then it won't show up in the ToC.

sarangan12:
@riknoll @rakatyal @dblotsky Removed the full examples for all except required.

www/static/js/docs.js
riknoll: Confusing for who? I don't think users expect the TOC to be that comprehensive; major headings only should be fine. We do need to be vigilant when we write pages to make sure the TOC displays correctly. I would argue that defining/using other tags just complicates the markdown and defeats the purpose of using it.

sarangan12: @dblotsky Its a tradeoff between the semantic meaning and what you want to show in ToC. And, this seems reasonable

sarangan12: @riknoll Ideally, I would have preferred a seperate commit. But this is a one-off case. With the changes to the events page, ToC will be cuttered. So, adding it here.

www/docs/en/dev/cordova/events/events.md
```diff
@@ -22,29 +22,386 @@ title: Events
...
+
+### Quick Example
+
+    document.addEventListener("pause", onPause, false);
+
+    function onPause() {
+        // Handle the pause event
+    }
+
+### Full Example
```
riknoll: The pause example remains

nikhilkh:
I'm assuming your grep'd for any incoming links to the files that you are deleting.

www/docs/en/dev/cordova/events/events.md
```diff
@@ -22,29 +22,384 @@ title: Events
...
+### Quick Example
+
+    document.addEventListener("volumeupbutton", onVolumeUpKeyDown, false);
+
+    function onVolumeUpKeyDown() {
+        // Handle the volume up button
+    }
+
+[UIApplicationExitsOnSuspend]: http://developer.apple.com/library/ios/#documentation/general/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html
+[AndroidLifeCycleGuide]: {{ site.baseurl }}/docs/en/dev/guide/platforms/android/lifecycle.html
```
nikhilkh: Should this be a relative reference instead?

PR #481, opened 2016-02-02
CB-10516: Remove h3-h6 from showing up in docs ToC

sarangan12:
Removing h3-h6 from showing up in docs ToC

@dblotsky @riknoll @nikhilkh Can you please review and merge this PR?

### DIFF
--- a/www/static/js/docs.js
+++ b/www/static/js/docs.js
@@ -40,7 +40,7 @@ $(document).ready(function () {
 
     // Table of Contents
     $('#page-toc').toc({
-        'selectors':         'h1,h2,h3,h4,h5,h6', // elements to use as headings
+        'selectors':         'h1,h2', // elements to use as headings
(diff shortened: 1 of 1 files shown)


riknoll:
LGTM

dblotsky:
I'm still a bit unconvinced that all H3 - H6 headings should be excluded from ToC across the board. This is specific to the reference pages, isn't it?

riknoll:
No, I don't think it is. What special meaning does H3+ have in the TOC? All of our TOC entries for a page are presented without hierarchy in the sidebar; H3+ lose all meaning in that format and become semantically the same in importance as H2 in the TOC, which does not make sense.

dblotsky:
Hmm, that's a good point. I guess if we ever do implement levels in the page ToC then we can always just put them back. LGTM then.

PR #487, opened 2016-02-05
Doc Anchor Fixes

dblotsky:
- Moving anchor creation out of ToC logic. 
- Moving ID creation into JS because RedCarpet doesn't support Unicode headings. 
- Making permalink icons a bit smaller and making them always show on small and medium screens.

### DIFF
--- a/www/static/js/docs.js
+++ b/www/static/js/docs.js
@@ -17,49 +17,82 @@
 
 $(document).ready(function () {
 
-    function getAnchorName(i, heading, prefix) {
-        var name = prefix;
-        if (heading.id) {
-            name += heading.id;
-        } else if (heading.name) {
-            name += heading.name;
-        } else {
-            name += i;
-        }
-        return name;
+    function slugify(text) {
+        text = text.toLowerCase();
+
+        // replace unaccepted characters with spaces
+        // NOTE:
+        //      a better regex would have been /[^\d\s\w]/ug, but the 'u' flag
+        //      (Unicode) is not supported in some browsers, and we support
+        //      many languages that use Unicode characters
+        text = text.replace(/[\[\]\(\)\=\+\?]/g, ' ');
+
+        // trim whitespace and replace runs of whitespace with single dashes
+        text = text.trim().replace(/ +/g, '-');
+
+        return text;
     }
 
-    var anchorLink = function (id) {
-        var anchor = document.createElement("a");
+    function getIdForHeading(heading) {
+        if (heading.id) {
+            return heading.id;
+        } else if (heading.name) {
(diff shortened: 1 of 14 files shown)


dblotsky:
@nikhilkh @riknoll @sarangan12 Please review when you folks get a chance.

www/static/js/docs.js
```diff
@@ -17,49 +17,81 @@
...
 
+    function anchorFor(id) {
+        var anchor       = document.createElement("a");
+        anchor.className = "fragment-anchor";
+        anchor.id        = id;
+        return anchor;
+    }
+
+    // go through all headings in the content and add some links
+    $('#page-toc-source h1, #page-toc-source h2').each(function (i, heading) {
```
nikhilkh: Does this still couple it to the TOC - `#page-toc-source`? Also, we should add links for all headings h3+ should be included.

www/static/js/docs.js
```diff
@@ -17,49 +17,81 @@
...
+            $(heading).append(permalinkTo(headingId));
+        }
+    });
+
     // Table of Contents
     $('#page-toc').toc({
         'selectors':         'h1,h2', // elements to use as headings
         'container':         '#page-toc-source', // element to find all selectors in
-        'prefix':            'link-', // prefix for anchor tags and class names
+        'prefix':            'toc-', // prefix for anchor tags and class names
```
nikhilkh: can we remove the prefix?

www/static/js/docs.js
riknoll: +1 The Android guide definitely requires referencing H3+

nikhilkh:
I'm assuming you grep'd for any `#link-` anchor links on our site that need to be changed now as we change the slugify scheme and prefix.

www/static/js/docs.js
dblotsky: Yep, I removed it from the code. I'll make it empty to make that more clear.

www/static/js/docs.js
dblotsky: Added H3.

dblotsky:
@nikhilkh: good catch! Fixed links with `#link-` in them.

dblotsky:
@nikhilkh I kept `#page-toc-source` because it's just an ID, which just happens to be used by the ToC too. If the ID ever changes or is moved, all uses of it should be examined and then we can create another one if required. The motivation to not touch it unless necessary is the fact that changing it will lead to a massive SVN commit (and right now it's not strictly necessary).

nikhilkh:
LGTM. Thanks, Dmitry for fixing this! I just tested it and it works very well!

PR #496, opened 2016-02-10
CB-10521 Add OS X platform to docs

tripodsan:


riknoll:
Hello again @tripodsan! There is currently an effort going on to reorganize a lot of stuff in the docs to fit the new [information architecture](https://github.com/cordova/cordova-discuss/pull/31) and one of the things that is changing is that all of the platform specific config.xml specifications are being combined into one reference. The [dev version of the config reference](https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/config_ref/index.md) has been updated to include all of the config.xml preferences for all of the platforms. It would be great to see the preferences you give here make their way into the main reference page. We have formatting for XML references defined in the [docs styleguide](https://github.com/apache/cordova-docs/blob/master/doc/docs-STYLEGUIDE.md#xml-references)

tripodsan:
thank's @riknoll, I will incorporate the changes. one question: why don't we include the version of the platform when the attribute/feature was included? I think this could be pretty useful for developers to figure out, why a certain attribute doesn't work...

riknoll:
Good point! @rakatyal do we have a standard way of doing that? Elsewhere in the docs we generally just mention it in the text and make the version bold, I think.

rakatyal:
Yes. We don't have a standard way of doing it for now. For now it's okay to specify it in bold.

tripodsan:
ok, I added the OS X specific preferences in the config reference.

riknoll:
I like how you put the version in the platform tag. It makes it clear without cluttering up the description by specifying the versions for several platforms. @rakatyal we should add something about versions to the styleguide, I think that it's a valid concern.

www/docs/en/dev/config_ref/index.md
```diff
@@ -58,7 +58,7 @@ platform. See [Icons and Splash Screens](images.html) for more information.
    id(string) | *Required* <br/> Specifies the app's reverse-domain identifier, and the `version` its full version number expressed in major/minor/patch notation.
    version(string) | *Required* <br/> Full version number expressed in major/minor/patch notation.
    versionCode(string) | ==Android== <br/> Alternative version for Android. For further details, see [Android versioning](http://developer.android.com/tools/publishing/versioning.html)
-   CFBundleVersion(string) |  ==iOS== <br/> Alternative version for iOS. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
+   CFBundleVersion(string) |  ==iOS== ==OS X== <br/> Alternative version for iOS and OS X. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
```
omefire: 'OS X' => 'OSX'

omefire:
@tripodsan Good Idea on adding OSX platform to the docs

omefire:
Missing preference: ```osx_CFBundleIdentifier```

tripodsan:
> Missing preference: `osx_CFBundleIdentifier`

where?

www/docs/en/dev/config_ref/index.md
```diff
@@ -320,8 +330,8 @@ platform. See [Icons and Splash Screens](images.html) for more information.
...
-   value(string or boolean) | *Required* <br/> ==iOS== ==Android== <br/> Specifies the name of the package to be used to initialize the plugin code (when the 'name' attribute is android-package or ios-package), specifies the name of the plugin to be loaded during controller initialization (when 'name' attribute is set to 'onload').
+   name(string) | *Required* <br/> Allowed values: android-package, ios-package, onload. <br/> ==iOS== ==OS X== ==Android== <br/> 'ios-package' and 'android-package' are used to specify the name of the package (as specified by the 'value' attribute) to be used to initialize the plugin code, while 'onload' is used to specify whether the corresponding plugin (as specified in the 'value' attribute) is to be instantiated when the controller is initialized.
```
omefire: Missing: ```osx-package```: 
https://github.com/apache/cordova-osx/blob/80c0c6fb970c26acaa3abf62200154a30dc634d0/CordovaLib/CordovaLib/Classes/Commands/CDVConfigParser.m#L77

tripodsan: yeah, both work: https://github.com/apache/cordova-osx/blob/80c0c6fb970c26acaa3abf62200154a30dc634d0/CordovaLib/CordovaLib/Classes/Commands/CDVConfigParser.m#L77, but I add it anyways.

tripodsan: thinking of this...maybe we should be more strict. and only allow `osx-package` in the future. created CB-10602

omefire:
>> Missing preference: osx_CFBundleIdentifier
> where?

Here: https://github.com/apache/cordova-osx/blob/cbfa7a531bb2a0a0e39fda14edaaff0e4a146b63/bin/templates/scripts/cordova/lib/prepare.js#L131

tripodsan:
but where is this documented? I can't even find where the ios_CFBundleIdentifier would be documented.
but there seems to be a mismatch in the osx parser: https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/metadata/osx_parser.js#L71

I created CB-10603 to investigate.

omefire:
I don't think ios_CFBundleIdentifier is documented in the source code. You can find details about it here:
- https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
- http://salesforce.stackexchange.com/questions/31235/where-can-i-find-my-bundle-identifier-and-why-is-it-important

tripodsan:
hmmm..ok...but then I don't understand your comment above:
> Missing preference: osx_CFBundleIdentifier

omefire:
> hmmm..ok...but then I don't understand your comment above:

> Missing preference: osx_CFBundleIdentifier

What I'm saying is that we should add it to the list of preferences, just like we've added ```DefaultVolumeStream```

omefire:
But I'm realizing that there is a mistake in the current docs. we should:
- have two entries: osx_CFBundleIdentifier and ios_CFBundleIdentifier
- remove CFBundleIdentifier

tripodsan:
ah, right: here it's still correct:
http://cordova.apache.org/docs/en/5.4.0/config_ref/index.html#additional-versioning

I'll update the config in my PR

www/docs/en/dev/config_ref/index.md
```diff
@@ -58,7 +58,8 @@ platform. See [Icons and Splash Screens](images.html) for more information.
...
    version(string) | *Required* <br/> Full version number expressed in major/minor/patch notation.
    versionCode(string) | ==Android== <br/> Alternative version for Android. For further details, see [Android versioning](http://developer.android.com/tools/publishing/versioning.html)
-   CFBundleVersion(string) |  ==iOS== <br/> Alternative version for iOS. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
+   ios-CFBundleVersion(string) |  ==iOS== <br/> Alternative version for iOS. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
+   osx-CFBundleVersion(string) |  ==OS X== <br/> Alternative version for OS X. For further details, see [OS X versioning](https://developer.apple.com/library/prerelease/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
```
dblotsky: @tripodsan Why do these need to be two different settings in `config.xml`? The linked Apple docs have identical descriptions.

www/docs/en/dev/config_ref/index.md
```diff
@@ -216,11 +220,13 @@ platform. See [Icons and Splash Screens](images.html) for more information.
...
    DisallowOverscroll(boolean) | *Default: false* <br/> ==iOS== ==Android== <br/> Set to **true** if you don't want the interface to display any feedback when users scroll past the beginning or end of content. On iOS, overscroll gestures cause content to bounce back to its original position. on Android, they produce a more subtle glowing effect along the top or bottom edge of the content. <br/>
    EnableViewportScale(boolean) | *Default: false* <br/>  ==iOS== <br/> Set to true to allow a viewport meta tag to either disable or restrict the range of user scaling, which is enabled by default. Place a viewport such as the following in the HTML to disable scaling and fit content flexibly within the rendering WebView: <br/> ```<meta name='viewport' content='width=device-width, initial-scale=1, user-scalable=no' />```
+   EnableWebGL(boolean) | *Default: false* <br/>  ==OS X **4.0.0**== <br/> Set to true to enable WebGL on the web view.
```
dblotsky: Is this not supported by other platforms?

www/docs/en/dev/config_ref/index.md
tripodsan: cordova-ios uses `ios-CFBundleVersion` (which is wrong in the current doc), so I thought it would be weird to use the same for osx. I'm all for having the same attribute for both platforms.

www/docs/en/dev/config_ref/index.md
tripodsan: this is an OSX WebView preference. I'm not sure about others: http://stackoverflow.com/questions/13659133/enable-webgl-in-the-mac-os-application-on-webkit

www/docs/en/dev/config_ref/index.md
dblotsky: @tripodsan is it hard to rename `ios-CFBundleVersion` to `CFBundleVersion`, and to have `ios-CFBundleVersion` just be an alias for it?

tripodsan: no, it's not hard. but we should discuss this on the list. `ios-CFBundleVersion` and `ios-CFBundleIdentifier` are the current used ones:
- https://github.com/apache/cordova-lib/blob/6.0.0/cordova-common/src/ConfigParser/ConfigParser.js#L106
- https://github.com/apache/cordova-lib/blob/6.0.0/cordova-common/src/ConfigParser/ConfigParser.js#L132

PR #512, opened 2016-02-22
Making windows guide single page

rakatyal:


riknoll:
LGTM

www/_data/redirects.yml
```diff
@@ -13,6 +13,8 @@ docs:
     - {old: "dev/guide/platforms/ios/config.html", new: "dev/config_ref/index.html"}
     - {old: "latest/guide/platforms/ios/tools.html", new: "latest/guide/platforms/ios/index.html"}
     - {old: "latest/guide/platforms/ios/config.html", new: "latest/config_ref/index.html"}
+    - {old: "dev/guide/platforms/win8/packaging.html", new: "dev/guide/platforms/win8/index.html#signing-an-app"}
+    - {old: "dev/guide/platforms/win8/win10-support.html", new: "dev/guide/platforms/win8/index.html"}
```
dblotsky: @rakatyal are these redirects for `dev` or `latest`? If the links to `dev` never went out, then the redirects should be for `latest`.

rakatyal: By went out you mean published to the site?

dblotsky: Yeah, either published to the site or printed in a blog post. My understanding is that the current `dev` redirects are there because there was a blog post about Android stuff that linked to `dev`.

rakatyal: Okay. The packaging one has been live since long so I will keep that one. Win10 support didn't went out so I will change it to latest.

dblotsky: For what it's worth, you can just add both.

dblotsky: @rakatyal: @r@riknoll just pointed out to me that adding `/latest/` redirects before we do the whole `6.x` thing will kill pages in `6.0.0`. Please then add the `/latest/` redirects commented-out.

PR #513, opened 2016-02-22
CB-10526: Temporarily remove redirects to latest until we are ready t…

omefire:
…o move the dev version of the docs to replace the latest version.

          They will be reinstated by CB-10672


omefire:
@riknoll to review.

riknoll:
LGTM

PR #514, opened 2016-02-22
Docs fetching improvements

dblotsky:
- Adding comments in `fetched-files.yml`
- Refactoring `fetch_docs.js`
- Removing `pluginapis.md` and placing plugins into the ToC
- Adding redirect for `pluginapis.md`


dblotsky:
@nikhilkh @sarangan12 @riknoll please review when you have a moment,

www/_data/toc/de-dev-manual.yml
```diff
@@ -1,96 +1,138 @@
...
-        -   name: "Hooks"
-            url: "guide/appdev/hooks/index.html"
-        -   name: "Plugin.xml"
-            url: "plugin_ref/spec.html"
+    -   name: "Config.xml"
+        url: "config_ref/index.html"
+    -   name: "Events"
+        url: "cordova/events/events.html"
+    -   name: "Plugin APIs"
+        children:
```
nikhilkh: It will be nice to put the plugin APIs at the end as the last reference topic - it will require less scrolling to get to the other topics.

nikhilkh:
LGTM. Good change! My commit depends on this change.

www/_data/toc/de-dev-manual.yml
dblotsky: Changed. I also reduced the padding in the ToC. Perhaps we should prioritize implementing the new ToC design, because it's getting a little unwieldy and the headings with children look really out of place.

tools/bin/fetch_docs.js
```diff
@@ -25,42 +25,103 @@ var child_process = require("child_process");
...
+        srcConfig.path = DEFAULT_REPO_PATH;
+    }
+
+    if (!srcConfig.commit) {
+        srcConfig.commit = getLatestRelease(srcConfig.packageName);
+    }
+
+    // complete dest config
+    if (!destConfig.permalink) {
+        destConfig.permalink = DEFAULT_PERMALINK_PREFIX + srcConfig.packageName + "/";
```
nikhilkh: We error our when there is no `destconfig` - however, there is a default behavior when the only property of destconfig not present. Either remove this default or remove failing on absence of `destconfig`

tools/bin/fetch_docs.js
```diff
@@ -25,42 +25,103 @@ var child_process = require("child_process");
...
+
+function getFetchedFile(entry) {
+
+    // get entry components
+    var srcConfig  = entry.src;
+    var destConfig = entry.dest;
+
+    // validate entry
+    if (!srcConfig) {
+        console.error("entry '" + entry.toString() + "' missing 'src'");
```
nikhilkh: Nit: Consider throwing an error object with appropriate message - throwing often has the nice side effect of presenting a stack. Also, the caller does not have to account for no return value.

tools/bin/fetch_docs.js
```diff
@@ -25,42 +25,103 @@ var child_process = require("child_process");
...
+    }
+
+    if (!destConfig) {
+        console.error("entry '" + entry.toString() + "' missing 'dest'");
+        return;
+    }
+
+    // complete src config
+    if (!srcConfig.packageName) {
+        srcConfig.packageName = packageNameFromRepoName(srcConfig.repoName);
```
nikhilkh: Nit: The JS way of writing this is: `srcConfig.packageName = srcConfig.packageName || packageNameFromRepoName(srcConfig.repoName);`
I like it because it's more compact too. This method seems long with a lot of whitespaces making it hard to eyeball - but all of this is just stylistic and not a big deal IMO.

www/_data/fetched-files.yml
```diff
@@ -1,153 +1,146 @@
...
     src:
-        repoName: apache/cordova-plugin-battery-status
-        packageName: cordova-plugin-battery-status
-        displayName: Battery Status
-        description: Monitor the status of the device's battery.
-    dest: cordova-plugin-battery-status/index.md
+        repoName: "apache/cordova-plugin-battery-status"
+        packageName: "cordova-plugin-battery-status"
+    dest:
+        permalink: "/docs/en/dev/cordova-plugin-battery-status/index.html"
```
nikhilkh: Nit: It might be nice to move all plugin docs into one directory "plugins"

tools/bin/fetch_docs.js
dblotsky: Yeah, it does seem paradoxical, but I made `dest` necessary because it could in the future have required arguments. I couldn't decide before, but you're right that it seems wacky. I will just make `permalink` mandatory as well.

tools/bin/fetch_docs.js
dblotsky: I opted for the "callee prints error and returns null, caller checks for null" instead of the "callee throws exception, caller catches it" pattern. Is one more advantageous than the other?

tools/bin/fetch_docs.js
dblotsky: I prefer very clear `if` statements because with long method/variable names sometimes the `||` is possible to miss. I opted for the former in this case because the names here are quite long.

www/_data/fetched-files.yml
dblotsky: Each string here is custom and totally up to change. For example the `/docs/en/dev/` part will have to go once we have translation. I figured the `plugin` part would be redundant since the word is already in the plugin's ID. I can change it to `plugin` though, if you'd prefer. I don't have much of a preference.

tools/bin/fetch_docs.js
nikhilkh: In this case you would not catch - it's a truly exceptional case and you can't handle it.

dblotsky: The current behavior is to ignore malformed entries and continue running.

PR #526, opened 2016-03-01
CB-10744 Fixing fragment link handling

dblotsky:
Adding support for name-based fragment anchors and anchors to IDs with special characters.

### DIFF
--- a/www/static/js/index.js
+++ b/www/static/js/index.js
@@ -145,17 +145,46 @@ $(document).ready(function () {
 
     // Smooth scroll to anchor links
     $("a[href^='#']").on('click', function(e) {
-        if(this.hash) {
+
+        // scroll only if there is a hash in the link's href
+        var hash = this.hash;
+        if (hash) {
 
             // prevent default anchor click behavior
             e.preventDefault();
 
-            // store hash
-            var hash = this.hash;
+            // get the fragment without the "#" symbol because location.hash
+            // is returned with it
+            var targetName = hash.slice(1);
+
+            // escape single quotes in target name because
+            // we use them in the selector to find matching targets
+            targetName.replace(/'/g, "\\\'").replace(/"/g, "\\\"");
+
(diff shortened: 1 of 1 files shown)


dblotsky:
@riknoll @nikhilkh please review when you have a moment.

www/static/js/index.js
```diff
@@ -145,17 +145,48 @@ $(document).ready(function () {
...
+        if (hash) {
 
             // prevent default anchor click behavior
             e.preventDefault();
 
-            // store hash
-            var hash = this.hash;
+            // get the fragment without the "#" symbol
+            var targetName = hash;
+            if (targetName[0] === "#") {
```
rakatyal: Do we need to return if this condition fails?

www/static/js/index.js
```diff
@@ -145,17 +145,48 @@ $(document).ready(function () {
...
+                var targetName = targetName.slice(1);
```
nikhilkh: We should not re-define targetName with a `var`

www/static/js/index.js
dblotsky: I guess the condition will never fail as long as browsers obey the standard. I'll remove the check.

rakatyal:
LGTM.

www/static/js/index.js
```diff
@@ -145,17 +145,46 @@ $(document).ready(function () {
...
+        if (hash) {
 
             // prevent default anchor click behavior
             e.preventDefault();
 
-            // store hash
-            var hash = this.hash;
+            // get the fragment without the "#" symbol because location.hash
+            // is returned with it
+            var targetName = targetName.slice(1);
```
riknoll: This line is a little confusing. Can you make the scoping of `targetName` clearer by renaming the variable?

dblotsky: Indeed. Turns out this line is also broken. Good catch! Fixed.

PR #532, opened 2016-03-02
Reverse-engineering GitHub's slugifying function.

dblotsky:


www/static/js/docs.js
```diff
@@ -17,19 +17,30 @@
...
+        var text = originalText;
+
+        // convert to lowercase
         text = text.toLowerCase();
 
-        // replace unaccepted characters with spaces
+        // replace whitespace characters with dashes
+        text = text.replace(/\s/g, '-');
+
+        // remove dashes at the start and end if they weren't there since the start
```
riknoll: If they weren't there from the start? What about these two cases:
```
"  -hello" => "--hello"
"   hello" => "hello"
```
Is that right? Should the first example work out that way with the extra dash? Seems like `trim()` should maybe get called before the first replace

dblotsky: These are the results from GitHub:

    "  -hello" => "---hello"
    "   hello" => "---hello"

That is, the truncation actually never happens. Fixing now.

PR #534, opened 2016-03-03
Documenting some missing prefs: WindowsStoreIdentityName, WindowsStor…

omefire:
…eDisplayName, WindowsStorePublisherName, BackgroundColor, WindowsToastCapable

### DIFF
--- a/www/docs/en/dev/config_ref/index.md
+++ b/www/docs/en/dev/config_ref/index.md
@@ -60,7 +60,7 @@ platform. See [Customize icons topic](images.html) for more information.
    android-versionCode(string) | ==Android== <br/> Alternative version for Android. Sets the [version code](http://developer.android.com/tools/publishing/versioning.html) for the application. See [the Android guide](../guide/platforms/android/index.html#setting-the-version-code) for information on how this attribute may be modified.
    ios-CFBundleVersion(string) |  ==iOS== <br/> Alternative version for iOS. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
    osx-CFBundleVersion(string) |  ==OS X== <br/> Alternative version for OS X. For further details, see [OS X versioning](https://developer.apple.com/library/prerelease/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
-   packageVersion(string) |   ==Windows== <br/> Alternative version for Windows. For futher details, see [Windows versioning](https://msdn.microsoft.com/en-us/library/windows/apps/br211441.aspx)
+   windows-packageVersion(string) |   ==Windows== <br/> Alternative version for Windows. For futher details, see [Windows versioning](https://msdn.microsoft.com/en-us/library/windows/apps/br211441.aspx)
    packageName(string) | *Default: Cordova.Example* <br/> ==Windows== <br/> Package name for Windows.
    xmlns(string) | *Required* <br/> Namespace for the config.xml document.
    xmlns:cdv(string) | *Required* <br/> Namespace prefix.
@@ -103,7 +103,7 @@ platform. See [Customize icons topic](images.html) for more information.
     <widget ...>
         <description>A sample Apache Cordova application</description>
     </widget>
-
+    
 
 ## author
    Specifies contact information that may appear within app-store lisitngs.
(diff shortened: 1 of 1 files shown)


www/docs/en/dev/config_ref/index.md
```diff
@@ -60,7 +60,7 @@ platform. See [Customize icons topic](images.html) for more information.
    versionCode(string) | ==Android== <br/> Alternative version for Android. For further details, see [Android versioning](http://developer.android.com/tools/publishing/versioning.html)
    ios-CFBundleVersion(string) |  ==iOS== <br/> Alternative version for iOS. For further details, see [iOS versioning](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
    osx-CFBundleVersion(string) |  ==OS X== <br/> Alternative version for OS X. For further details, see [OS X versioning](https://developer.apple.com/library/prerelease/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).
-   packageVersion(string) |   ==Windows== <br/> Alternative version for Windows. For futher details, see [Windows versioning](https://msdn.microsoft.com/en-us/library/windows/apps/br211441.aspx)
+   windows-packageVersion(string) |   ==Windows== <br/> Alternative version for Windows. For futher details, see [Windows versioning](https://msdn.microsoft.com/en-us/library/windows/apps/br211441.aspx)
```
nikhilkh: The example below also needs an update. Also, we should really have one `widget` element with the attribute - as that's how people should use this.

www/docs/en/dev/config_ref/index.md
```diff
@@ -211,9 +211,8 @@ platform. See [Customize icons topic](images.html) for more information.
...
-   AppendUserAgent(string) | ==Android== <br/> If set, the value will append to the end of old UserAgent of webview. When using with OverrideUserAgent, this value will be ignored.
-   AppendUserAgent(string) | ==iOS== <br/> If set, the value will append to the end of old UserAgent of webview. When using with OverrideUserAgent, this value will be ignored.
-   BackgroundColor(string) | ==Android== ==BlackBerry== ==Windows== <br/> Supports a four-byte hex value, with the first byte representing the alpha channel, and standard RGB values for the following three bytes.
+   AppendUserAgent(string) | ==Android== ==iOS== <br/> If set, the value will append to the end of old UserAgent of webview. When using with OverrideUserAgent, this value will be ignored.
+   BackgroundColor(string) | ==Android== ==BlackBerry== ==Windows== <br/> Supports a four-byte hex value, with the first byte representing the alpha channel, and standard RGB values for the following three bytes. <br/> For Windows, it is a three-byte hexadecimal number preceded by "#" or a named color.
```
nikhilkh: Should we comment that the alpha channel byte is dropped by Cordova windows build automatically?

www/docs/en/dev/config_ref/index.md
```diff
@@ -245,6 +244,9 @@ platform. See [Customize icons topic](images.html) for more information.
    PopupBlocker(string) | *Default: enable* <br/> ==BlackBerry== <br/> Enables the popup blocker, which prevents calls to window.open(). By default, popups display in a child browser window. Setting the preference to enable prevents it from displaying at all.
    SetFullscreen(boolean) | *Default: false* <br/> ==Android== <br/> Same as the Fullscreen parameter in the global configuration of this xml file. This Android-specific element is deprecated in favor of the global Fullscreen element, and will be removed in a future version.
    ShowTitle(boolean) | *Default: false* <br/> ==Android== <br/> Show the title at the top of the screen.
+   SplashScreen(string) | *Default: splash* <br/> ==Android== <br/> The name of the file minus its extension in the ```res/drawable``` directory. Various assets must share this common name in various subdirectories.
```
nikhilkh: We do not document SplashScreen config properties as they are considered part of the SplashScreen plugin - even though the code is part of each platform. We should probably mention one of them here without providing details (so as not to duplicate the docs) and say that splashscreen plugin is required for them to work and they are documented there!

www/docs/en/dev/config_ref/index.md
omefire: Good catch.

www/docs/en/dev/config_ref/index.md
omefire: Sure, didn't know about that.

www/docs/en/dev/config_ref/index.md
omefire: Since they won't work without the splashscreen plugin, It'd be better to omit them. 

The fact that their code is part of each platform instead of the plugin's native code (if I understood you correctly) is an implementation detail.

www/docs/en/dev/config_ref/index.md
omefire: Is this not covered by the current comment ? "For Windows, it is a 3-byte hexadecimal number..."

omefire:
@rakatyal

www/docs/en/dev/config_ref/index.md
nikhilkh: "For Windows, the alpha channel is ignored" is more clear than saying that value should be 3-byte for windows - as it's OK to have a 4-byte value for windows - but the alpha channel is not used. 

Also, the documentation does not tell me what the background color does altogether? What is this the background color for?

PR #535, opened 2016-03-03
Adding OS X platform guide to TOC

rakatyal:


riknoll:
I think you are missing some hyphens

riknoll:
LGTM

PR #551, opened 2016-03-10
CB-10836 Improving "Edit" links

dblotsky:
# Changes

- Added links to Crowdin translation on appropriate pages
- Changed "Edit" links for translated content to point to English sources
- Changed "Edit" links for current content to point to /dev/ sources
- Changed all other "Edit" links point to their respective files

# Examples (link at each example's top)

### DIFF
--- a/conf/_config.yml
+++ b/conf/_config.yml
@@ -27,6 +27,9 @@ latest_docs_version: dev
 # the docs version that is shown by default when clicking on docs links on the site
 default_linked_docs_version: latest
 
+# the language that serves as the source for the translations
+src_language: en
+
 # NOTE:
 #      this value _might_ get overwritten on some pages;
 #      see _defaults.yml, which is a generated file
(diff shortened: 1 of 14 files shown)


csantanapr:
@dblotsky Love the option to help with translations to get them to the right spot in Crowding.

Still I think the label for the button should be changed from "Edit" that was my initial concern. just Edit it's hard to spot and not knowing if it's going to login the user like the wiki

Can we have for English the button say "Edit Source on Github" and just be a Link, this way Link means you are going to leave this page and visit Gihub. Since there is one option.

Other Languages I would put two links "Edit Source on Gihub" and "Edit Translation on Crowding"

stevengill:
LGTM

dblotsky:
@csantanapr the links don't fit on mobile screens, which is why I went for a dropdown.

csantanapr:
@dblotsky no way we can make it work on mobile float, flex, css, etc.. ?

csantanapr:
They don't have to be on the row on mobile.

dblotsky:
@csantanapr please take a look at #555. It has the links implementation and a screenshot of what it looks like on mobile.

csantanapr:
@dblotsky  LGTM !

dblotsky:
@csantanapr which of the two implementations do you think we should go for?

csantanapr:
@dblotsky I vote for using links #555

dblotsky:
Closing in lieu of #555.

PR #552, opened 2016-03-10
Fix typo for orientation property portrait.

insane210:


dblotsky:
@insane210 thank you for the PR! :)

PR #560, opened 2016-03-18
Better syntax highlighting

riknoll:
Switching from prettify to rouge!

Good changes:
* Way nicer looking code blocks that we can easily edit the CSS for
* No more inline code highlighting (aka less colorful blog posts)
* Better performance probably?
* I fixed a lot of random broken markdown across the site

Bad changes:
* Language detection is no longer automatic so we need to use GFM fenced code blocks (which I already did for `en/dev/`)

I highly recommend that you add `?w=1` to the URL when viewing changes in this PR (removes whitespace changes)


riknoll:
@dblotsky to review (sorry)

riknoll:
JIRAs are incoming. Also, I added the license to the `syntax.css` file. It was obtained from here: https://github.com/mojombo/tpw/blob/master/css/syntax.css

dblotsky:
LGTM.

riknoll:
Jira is [CB-10907](https://issues.apache.org/jira/browse/CB-10907)

PR #563, opened 2016-03-19
Removing prettify reference in index.js

riknoll:
Forgot that we referenced a function from the Prettify script in another script that was included later in the page. Hurray web development!

@dblotsky check it out


dblotsky:
Whoop, good catch! LGTM.

PR #567, opened 2016-03-24
CB-11021 Changing docs styling

dblotsky:
- Styling ToC
- Changing color scheme for headings
- Changing the font

Result:
capture


dblotsky:
@nikhilkh @sarangan12 @riknoll @rakatyal: whenever any of you have any time, please give this a look.

riknoll:
Alright, on the font front Google provides the appropriate attribution notice here: https://www.google.com/fonts/attribution

The question is where should we put it? Maybe we can place  a comment in the source and add it to the README as well with a link to the font's page.

riknoll:
Otherwise, LGTM. The TOC logic is unchanged, right? Kind of hard to read through the diff on that. Tested it in Chrome and Edge and everything looks great. Gigantic improvement.

riknoll:
Also, create a JIRA!

www/_includes/footer.html
```diff
@@ -1,6 +1,6 @@
 <div class="blue-divider"></div>
 <footer>
-    <div class="container" id="footer">
```
rakatyal: Why this change? Will it effect non-docs pages as well?

dblotsky:
@riknoll thanks for pointing that out! I've created a JIRA: https://issues.apache.org/jira/browse/CB-11021.  I'll also add the font attribution to the footer and the README.

The ToC logic has slightly changed. The following cases used to be possible, and I removed them:
- setting the `url` property on entries that had the `children` property
- skipping entries that had an empty `url` property

Now all entries are printed and I just commented out the ones that don't have URIs, and the entries with children can no longer be links.

www/_includes/footer.html
dblotsky: Great catch! It makes the footer look a bit strange on other pages. I'll come up with a fix.

dblotsky:
@riknoll Please give this another look. I added an attributions page. @rakatyal I fixed the footer to be slightly different for the docs and for the rest of the site.

riknoll:
@dblotsky Is there a reason this suddenly has 1,634 changed files?

dblotsky:
@riknoll yeah. The latest merge to the ASF Git master hasn't reflected in the GitHub master yet because it was a large commit. Sorry, the best way to see the diff for now is probably to check out the branch and do the diff locally.

PR #569, opened 2016-03-25
Adding both Chinese versions.

dblotsky:
- Copied `zh` into `zh-tw` and added `zh-cn`
- Added redirect for `/zh/` to `/zh-tw/`
- Fixed the language name in `gen_versions.js`
- Fixed regexes to accept language codes of the form `XX` and `XX-YY`


dblotsky:
@riknoll @rakatyal please review when you get a chance.

dblotsky:
@rakatyal @riknoll ping.

www/.htaccess
```diff
@@ -18,29 +18,31 @@ RewriteEngine on
...
-#       these can both apply, turning "/docs" into "/docs/en/latest/"
-RewriteRule ^.*docs/$ {{site.baseurl}}/docs/en/
-RewriteRule ^.*docs/(\w\w)/$ {{site.baseurl}}/docs/$1/latest/ [R=302,L]
-
-# 302 (temp):
-#       docs/XX/latest -> docs/XX/latest/
-RewriteRule ^.*docs/(\w\w)/latest$ {{site.baseurl}}/docs/$1/latest/ [R=302,L]
+#       these can all apply, turning "/docs" into "/docs/[default language]/latest/"
+RewriteRule ^.*docs/$ {{site.baseurl}}/docs/{{site.language}}/
+RewriteRule ^.*docs/(\w\w(?:-\w\w)?)$ {{site.baseurl}}/docs/$1/
```
riknoll: What is this rule doing? Looks like it is just adding a slash to the end of the URL?

dblotsky: Yep, that's exactly what it's doing. This way `/docs/en` and `/docs/zh-tw` become `/docs/en/` and `/docs/zh-tw/` respectively. This is because `/docs/en` resolves to `/docs/en` (which doesn't exist), and `/docs/en/` resolves to `/docs/en/index.html`, which exists.

riknoll: `/docs/en/index.html` does not exist, right?

Looks like `/docs/en` will go through three redirects:
```
/docs/en -> /docs/en/ -> /docs/en/latest -> /docs/en/latest/
```

Can't all of those just be one step instead of chaining like that? Makes it hard to read.

dblotsky: Oh, you're right! That doesn't exist. I said `/docs/en/` but I was thinking about `/docs/en/latest/`.

The reason there is a chain is that any intermediary URI will "fall through" to `/docs/en/latest/`. So all of `/docs/en`, `/docs/en/`, `/docs/en/latest` will become `/docs/en/latest/`.

riknoll: Yeah! That's what I was suggesting

dblotsky: @riknoll I've made the change. How does it look now?

PR #574, opened 2016-04-08
CB-10968 Update iOS 'Requirements & Supports' section

omefire:


jcesarmobile:
Sorry, the list I provided was for devices that supported iOS 8. I think we should add the devices that went out with iOS 9 (iPhone 6S, iPhone 6S Plus, iPhone SE and iPad Pro)

omefire:
hmm, missed that earlier. Added them: https://en.wikipedia.org/wiki/IOS_9#Supported_devices

PR #577, opened 2016-04-12
CB-11064: Adding a note to the engine tag spec and fixing a header

riknoll:
@nikhilkh when you have a moment


purplecabbage:
lgtm!

nikhilkh:
LGTM

PR #580, opened 2016-04-18
CB-11077 backbutton does not minimize app

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-11077)

### DIFF
--- a/www/docs/en/dev/cordova/events/events.md
+++ b/www/docs/en/dev/cordova/events/events.md
@@ -324,6 +324,24 @@ function onBackKeyDown() {
 }
 ```
 
+### Windows Quirks
+
+Throw an error in a `backbutton` callback to force the default behavior, which is an app exit:
+
+```javascript
+document.addEventListener('backbutton', function (evt) {
+    if (cordova.platformId !== 'windows') {
+        return;
+    }
+
+    if (window.location.href !== firstPageUrl) {
+        window.history.back();
+    } else {
+        throw new Error('Exit'); // This will suspend the app
+    }
+}, false);
+```
+
 ## menubutton
 
 The event fires when the user presses the menu button. Applying an event handler


www/docs/en/dev/cordova/events/events.md
```diff
@@ -323,6 +323,16 @@ function onBackKeyDown() {
 }
 ```
 
+### Windows Quirks
+
+Throw an error in a `backbutton` callback to force the default behavior:
+
+```javascript
+document.addEventListener('backbutton', function (evt) {
+    throw "Exit"; // This will suspend the app
```
nikhilkh: Should it throw a new Error instead of a string?

www/docs/en/dev/cordova/events/events.md
```diff
@@ -323,6 +323,16 @@ function onBackKeyDown() {
...
+}, false);
+```
+
```
nikhilkh: Is there no better way to handle this?

www/docs/en/dev/cordova/events/events.md
```diff
@@ -323,6 +323,16 @@ function onBackKeyDown() {
 }
 ```
 
+### Windows Quirks
+
+Throw an error in a `backbutton` callback to force the default behavior:
```
nikhilkh: What is the `default` behavior here?

daserge: The default is to exit if there is no history.

www/docs/en/dev/cordova/events/events.md
daserge: The throw value is not used anywhere, it's just being caught in cordova-js and causes the event handler to indicate that the event has not been handled:
https://github.com/apache/cordova-windows/blob/master/template/www/cordova.js#L1489

www/docs/en/dev/cordova/events/events.md
nikhilkh: For a single page app, it will result in exiting? Shall we add that to the doc comment.

www/docs/en/dev/cordova/events/events.md
nikhilkh: Why override the backbutton then? Should not overriding result in the default behavior anyway?

www/docs/en/dev/cordova/events/events.md
daserge: I don't see another way to handle this situation taking into account the cordova-js implementation.
Theoretically we could use a callback return value from `cordova.fireDocumentEvent('backbutton', evt, true);` but it only allows to capture an exception but not a value.

www/docs/en/dev/cordova/events/events.md
daserge: I believe the goal of the overriding is to implement a custom navigation logic as back button closes the app by default on Windows as opposed to its purpose - go back or close if there is no navigational history.
So the only way to control this now is an option to throw an exception in handler so that the app close will happen.

www/docs/en/dev/cordova/events/events.md
daserge: I'll update the PR, thanks for the review.

beat-at-hartwig:
It is mandatory to implement this kind of backbutton handling by default due to  [Windows Store Policies](https://msdn.microsoft.com/en-us/library/windows/apps/dn764944.aspx#POL_10_4_4). My last app submission was declined because this bahaviour was not implemented. The problem occurs only on cordova-windows@4.2.0+ where back button handling was introduced.

daserge:
@beat-at-hartwig, you can handle both first page and sub page cases in `backbutton` event and do `throw` or `history.back();` correspondingly.
This can be used as a temporary measure but I agree that the behavior isn't consistent with Android now, which has the correct behavior out of the box.

beat-at-hartwig:
@daserge Thank you for your suggestions. I already implemeted similar code fitting my needs. Doing a `throw` was a nice hint. Thanks!

daserge:
@nikhilkh, I've updated the PR, please take a look once more.

PR #582, opened 2016-04-19
CB-11091 added plugins release blog post

stevengill:


riknoll:
It might be helpful to mention that cordova-plugin-compat is replacing the PermissionHelper.java file that was duplicated in all the plugin repos. There were a few plugin authors who have already copied that file from the core plugins and they should probably update to using cordova-plugin-compat.

rakatyal:
LGTM.

PR #584, opened 2016-04-22
CB-11127 Improving internal linking

dblotsky:
Relevant changes:
- added a new data file, `all-pages.yml`, which stores a list of valid links on the site
- version dropdowns and out-of-date alerts now use the new data file to only link to valid pages
- redirects are now a dictionary instead of a list, which is more flexible and easier to work with

Miscellaneous changes:
- refactored Makefile dependencies to properly rebuild some files when necessary
- refactored `docs.html` to be a bit more readable

### DIFF
--- a/tools/bin/gen_pages_dict.js
+++ b/tools/bin/gen_pages_dict.js
@@ -1,0 +1,120 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+"use strict";
+
+var fs       = require("fs");
+var path     = require("path");
+var yaml     = require("js-yaml");
+var walk     = require("walk");
+var glob     = require("glob");
+var Q        = require("q");
+var optimist = require("optimist");
+
+var util = require("./util");
+
+// constants
+var LATEST_ALIAS_URI = "/latest/";
+
+// helpers
+function pathToURI(filePath, rootPath) {
+    return filePath
+        .replace(new RegExp("^" + rootPath), "")
+        .replace(new RegExp("\\.md$"), ".html");
+}
+
+function pagesFromRedirects(redirects, languages) {
+    var pages = {};
+
+    // add docs redirects
+    if (typeof redirects.docs !== "undefined") {
+        for (var redirectSource in redirects.docs) {
+
+            // add an entry for the redirect's source, once for each language
+            for (var i = 0; i < languages.length; i++) {
+                var language = languages[i];
+                var pagePath = "/docs/" + language + "/" + redirectSource;
+
+                pages[pagePath] = true;
+            }
+        }
+    }
+
+    return pages;
+}
+
+function isInLatestDocs(uri, latestVersion) {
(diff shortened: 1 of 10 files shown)


tools/bin/gen_pages_dict.js
```diff
@@ -0,0 +1,120 @@
...
+        for (var i = 0; i < filePaths.length; i++) {
+            var filePath = filePaths[i];
+            var fileURI  = pathToURI(filePath, siteRootPath);
+
+            // add the page
+            pages[fileURI] = true;
+
+            // also add /latest/ version for pages in latest docs
+            if (isInLatestDocs(fileURI, latestVersion)) {
+                var latestURI = fileURI.replace("/" + latestVersion + "/", LATEST_ALIAS_URI);
```
sarangan12: LATEST_ALIAS_URI has a '/' in the beginning. Here I see that "/" is added before LATEST_ALIAS_URI. So, would that be changing the path as **//**latest/. Is this what we want?

www/.htaccess
```diff
@@ -61,14 +63,14 @@ RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/latest$ {{site.baseurl}}/docs/$1/latest/ [R
 #
 #     docs/*/XX/* -> docs/*/YY/*
 #
-{% for redirect in site.data.redirects.version-renames %}RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/{{redirect.old}}/(.*)$ {{site.baseurl}}/docs/$1/{{redirect.new}}/$2 [R=302,L]
+{% for redirect in site.data.redirects.version-renames %}RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/{{redirect[0]}}/(.*)$ {{site.baseurl}}/docs/$1/{{redirect[1]}}/$2 [R=302,L]
```
sarangan12: NitPick: Alignment. Whouldn't this be better 

{% for redirect in site.data.redirects.version-renames %}
               RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/{{redirect[0]}}/(.*)$ {{site.baseurl}}/docs/$1/{{redirect[1]}}/$2 [R=302,L]

www/.htaccess
```diff
@@ -61,14 +63,14 @@ RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/latest$ {{site.baseurl}}/docs/$1/latest/ [R
...
 {% endfor %}
 
 # 302 (temporary):
 #
 #     docs/XX/* -> docs/YY/*
 #
-{% for redirect in site.data.redirects.language-renames %}RewriteRule ^.*docs/{{redirect.old}}/((?:\d+\.(?:\d+\.\d+|x))|dev|latest)/(.*)$ {{site.baseurl}}/docs/{{redirect.new}}/$1/$2 [R=302,L]
+{% for redirect in site.data.redirects.language-renames %}RewriteRule ^.*docs/{{redirect[0]}}/((?:\d+\.(?:\d+\.\d+|x))|dev|latest)/(.*)$ {{site.baseurl}}/docs/{{redirect[1]}}/$1/$2 [R=302,L]
```
sarangan12: Nitpick: Alignment

dblotsky:
@riknoll @rakatyal please also give this a look when you have a moment.

www/.htaccess
dblotsky: Alignment with what?

tools/bin/gen_pages_dict.js
dblotsky: This replaces `/6.x/` with `/latest/`, so it accounts for the slashes.

sarangan12:
Got it.... LGTM

www/.htaccess
dblotsky: The resulting `.htaccess` file has too many spaces if we add an extra space like this:

    {% for ... %}
    RewriteRule ...
    {% endfor %}

because it will render as:

    RewriteRule ...
    
    RewriteRule ...

We can still refactor it from this:

    {% for ... %}RewriteRule ... 
    {% endfor %}

to this:

    {% for ... %}
    RewriteRule ... {% endfor %}

Which do you think is better?

www/.htaccess
dblotsky: Oh, sorry. Just saw your comment before.

riknoll:
LGTM

PR #585, opened 2016-04-26
Use XML entities in attributes

nikhilkh:
This makes the XML well formed


purplecabbage:
These were rendering properly in github's markdown, is this an issue with our own document md->html conversion?

image

The previous version was passing online xml validation tools.
I worry that if we change the docs to be 'proper', developers will use this syntax verbatim and our own code in cordova-lib will not handle it properly. Do you know if we handle this &gt; and > equally?

nikhilkh:
It's really an issue with some XML parsers that conform to the spec on "well-formed" XML. In particular, .NET XML parsers used by Visual Studio fail if you do not use entities within attributes.

Our own node code uses element tree which works well with entities in attributes.

PR #592, opened 2016-05-03
Improving SEO for plugin api pages

riknoll:
This changes all of our canonical links in docs to point to the pages at latest if available. It also adds redirects for all old versions of plugin pages to the current READMEs so that they can use the canonical link improvements as well. As a bonus, the "click here for latest" link will work for those pages too.

One caveat: non-English URLs will get 404s when going to latest. We need to copy over the READMEs for languages other than English, which should hopefully be done when we pull from crowdin next.

I might need to test this before deploying. It's tough to test redirects, though.


riknoll:
@nikhilkh @rakatyal take a look

rakatyal:
LGTM.

nikhilkh:
👍

PR #593, opened 2016-05-03
CB-11196: Converting mark elements in docs to platform logos

riknoll:
Changes the XML references to display platform logos rather than lists of highlighted platform names. Also adds new logos for iOS, OS X, and Browser platforms that will appear both in the docs and in plugin search.


riknoll:
Here is what it looks like. The logos appear under the attribute name.

new_logos_xml

riknoll:
Also note that this changes the iOS logo (see leftmost logo in attached image). The Apple logo is now used to indicate the OS X platform. The new iOS logo is sort of hard to see in the xml docs, but looks much better in the plugin search. All of the logos have alt-text that says the platform they represent.

jasongin:
> The Apple logo is now used to indicate the OS X platform.
This is confusing to me, given that I never think about using Cordova to develop apps for OS X, and my iPhone has a prominent Apple logo on the back.

jasongin:
Do the icons have title text? If so I can't find it in the code.

purplecabbage:
Diffing an svg file also sucks.

riknoll:
@jasongin maybe I can see about getting [this logo](http://www.apple.com/osx/) for OS X.

Title text is added by docs.js [here](https://github.com/apache/cordova-docs/pull/593/commits/2db61d31e3b73733449a075d0330dd4f781de0e2#diff-ab90a31f2984d41d3f15121f790a71deR138). I hate doing it in the JS, but that's the only way if we ant to use the `== mark ==` syntax

riknoll:
@jasongin @rakatyal @purplecabbage updated. iOS logo back to normal and new logo for OS X.

new_logos_xml_2

new_logos_plugin

jasongin:
👍

nikhilkh:
Wohoo! 💥

rakatyal:
👍

PR #595, opened 2016-05-05
CB-11203: Improving documentation of plugin CLI variables

riknoll:
I also uncovered [CB-11205](https://issues.apache.org/jira/browse/CB-11205) while documenting this.

### DIFF
--- a/www/docs/en/dev/config_ref/index.md
+++ b/www/docs/en/dev/config_ref/index.md
@@ -220,7 +220,10 @@ platform. See [Customize icons topic](images.html) for more information.
    ```
 
 ## plugin
-   Specifies details about what plugin to restore during a prepare.
+   Specifies details about what plugin to restore during a prepare. This element
+   is automatically added to a project's `config.xml` when a plugin is added using
+   the `--save` flag. See the [CLI reference][plugin_cli] for more information on
+   adding plugins.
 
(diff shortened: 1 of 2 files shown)


riknoll:
@nikhilkh please review.

www/docs/en/dev/config_ref/index.md
```diff
@@ -234,6 +234,23 @@ platform. See [Customize icons topic](images.html) for more information.
    <plugin name="cordova-plugin-device" spec="https://github.com/apache/cordova-plugin-device.git#1.0.0" />
    ```
 
+### variable
+   Persists the value of a CLI variable to be used when restoring a plugin during a prepare. See the [preference element][plugin_preference]
```
nikhilkh: It will be great to specify that edits to the value in config.xml do not take effect without removing and adding the plugin. This is added by --save option in the CLI.

nikhilkh:
:+1:

riknoll:
@nikhilkh updated. I also added pointers to the CLI reference for the `--save` flag. Is that the right place? We are still planning on getting rid of [this page](http://cordova.apache.org/docs/en/latest/platform_plugin_versioning_ref/index.html) right?

nikhilkh:
LGTM. Yes, it's a goal to remove that one.

PR #596, opened 2016-05-09
Moving analytics script to head

riknoll:
It looks like it got moved in a refactor. It needs to be in the head of every page to work properly according to this: https://support.google.com/analytics/answer/1008080?hl=en&ref_topic=1008079


nikhilkh:
LGTM

PR #598, opened 2016-05-10
Fixing markdown for config.xml

rakatyal:


rakatyal:
@riknoll: Please review.

PR #600, opened 2016-05-10
Update SQLite plugin details

nikhilkh:
@brodybits to help review.

### DIFF
--- a/www/docs/en/dev/cordova/storage/storage.md
+++ b/www/docs/en/dev/cordova/storage/storage.md
@@ -294,8 +294,7 @@ The main differences are:
 
 It is available in the following variations:
 
-* **[cordova-sqlite-storage][SQLiteStorage]** - core version that relies
-  on native SQLite implementation. As such, it supports iOS, Android & Windows platforms.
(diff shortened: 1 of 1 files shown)


brodybits:
I don't think "relies on native SQLite implementation" is so relevant anymore. The cordova-sqlite-storage version now includes its own sqlite3 implementation. Also yours truly (@brodybits) broke the cordova-sqlite-storage README link.

For the README links, would it be an idea to consider using "#readme" at the end of the repo link instead of hardcoding the link through the default branch name?

FYI other changes may be coming:
- An "express-core" version  that would use the built-in sqlite libraries for Android and iOS (not sure if we should really link this)
- The default branch of the SQLCipher version (for which there is no link) will probably change
- I may move or rename the "evfree" version.

In the past I would maintain the plugin source code in separate branches and publish a "*-master" branch that would also include the sqlite3 source/library dependencies. I changed cordova-sqlite-storage and cordova-sqlite-ext to fetch the sqlite3 source/library dependencies from npm (as discussed in cordova/cordova-discuss#38) and published them from lighter-weight branches. I expect to do the same for the SQLCipher and "evfree" versions.

brodybits:
The updates look good to me, thanks!

FYI I dropped the Cordova engine constraint on Cordova-sqlite-storage.

PR #608, opened 2016-06-02
Update platform/plugin versioning

carynbear:
Remove cordova plugin update (this command does not seem to exist)

### DIFF
--- a/www/docs/en/dev/platform_plugin_versioning_ref/index.md
+++ b/www/docs/en/dev/platform_plugin_versioning_ref/index.md
@@ -144,18 +144,15 @@ $ cordova plugin save
 ```
 
 
-### Updating / Removing plugins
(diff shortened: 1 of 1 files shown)


stevengill:
Thanks @carynbear! I don't think we ever supported a plugin update command.

fredgalvao:
Why not implement it instead? I'd  bet `cordova plugin rm XXX && cordova plugin add XXX` is a very common thing, and a `cordova plugin update XXX` could be usefull.

I've had situations where I couldn't update a single plugin with that workaround because of dependencies, so an `update` command wouldn't be a shortcut only.

stevengill:
@fredgalvao, I agree that there would be benefits to implementing it. Feel free to open a feature request issue at issues.cordova.io. 

Regardless, this PR is to fix the docs issue.

PR #612, opened 2016-06-17
CB-11412 Added docs for template use and create

carynbear:


www/_data/redirects.yml
```diff
@@ -348,6 +348,7 @@ docs-global:
     "guide_appdev_privacy_index.md.html": "guide/appdev/privacy/index.html"
     "guide_appdev_whitelist_index.md.html": "guide/appdev/whitelist/index.html"
     "guide_cli_index.md.html": "guide/cli/index.html"
+    "guide_cli_template.md.html": "guide/cli/template.html"
```
dblotsky: Did this link exist in the past? If not, the redirect isn't necessary.

PR #614, opened 2016-06-23
CB-11023 New edit-config tag in plugin.xml

ktop:
Documenting the new edit-config tag in plugin.xml. 
@riknoll can you review? I tried to keep the examples short, but explain enough to get the point across. Let me know if I need to fix anything.

### DIFF
--- a/www/docs/en/dev/plugin_ref/spec.md
+++ b/www/docs/en/dev/plugin_ref/spec.md
@@ -365,6 +365,112 @@ For windows-specific attributes:
 ```
 The above example will set pre-8.1 platforms (Windows 8, specifically) to require the `webcam` device capability and the `picturesLibrary` general capability, and apply the `webcam` device capability only to Windows 8.1 projects that build for Windows Phone.  Windows desktop 8.1 systems are unmodified.
 
+### edit-config
+Similar to `config-file`, `edit-config` identifies an XML-based configuration file to be modified, where in that document the modification should take place, and what should be modified. Instead of appending new children to an XML document tree, `edit-config` makes modifications to attributes of XML elements. There are two modes which will determine what type of attribute modification will be made, `merge` or `overwrite`. `edit-config` has one child and that child will contain the attributes to be added.
+
+Attributes(type) <br/> <span class="sub-header">Only for platform:</span> | Description
+---------------- | ------------
+file(string) | The file to be modified, and the path relative to the root of the Cordova project. If the specified file does not exist, the tool ignores the configuration change and continues installation. <br/> The target can include wildcard (`*`) elements. In this case, the CLI recursively searches through the project directory structure and uses the first match. <br/> On iOS, the location of configuration files relative to the project directory root is not known, so specifying a target of `config.xml` resolves to `cordova-ios-project/MyAppName/config.xml`.
+target(string) | An XPath selector referencing the target element to make attribute modifications to. If you use absolute selectors, you can use a wildcard (`*`) to specify the root element, e.g., `/*/plugins`. If the selector does not resolve to a child of the specified document, the tool stops and reverses the installation process, issues a warning, and exits with a non-zero code.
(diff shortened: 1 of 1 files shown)


macdonst:
@ktop looks good but my personal preference would be to show a before and after of the Android Manifest.xml you use in your example to really hammer home how it works.

riknoll:
For the examples, I would just give a snippet of the AndroidMainfest and not the whole thing. There's a lot that doesn't change in the before and after. Secondly, the example of force adding plugins is pretty good, but I wonder if this is the right place for it. Perhaps it is better suited for the CLI guide as a sub-heading under the [plugin command](https://github.com/apache/cordova-cli/blob/master/doc/readme.md#cordova-plugin-command). People who get CLI errors might be more likely to look there since they are probably not plugin authors themselves.

ktop:
I shortened the AndroidManifest to snippets. 

I could see the force add content going into both locations, but in the CLI guide it should be more describing the error about conflicting edit-config and how it is caused. Maybe this section can link to the plugin.xml guide about edit-config for how to resolve the error. It's probably worth it for users to learn about edit-config in the plugin.xml guide before making changes.

www/docs/en/dev/plugin_ref/spec.md
```diff
@@ -365,6 +365,111 @@ For windows-specific attributes:
...
+    ...
+        <activity android:configChanges="orientation|keyboardHidden" android:label="NewLabel" android:name="MainActivity">
+            ...
+        </activity>
+    ...
+    <uses-sdk android:maxSdkVersion="23" android:minSdkVersion="16" android:targetSdkVersion="23" />
+</manifest>
+```
+
+Multiple plugins will not be able to to modify the same attributes because it may cause problems with the application. An error will be thrown and plugin install will fail. The conflicting `edit-config` tags must be resolved before the plugin can be added. Make modifications to conflicting tags to resolve the conflict, then remove and re-add those updated plugins.
```
riknoll: "Multiple plugins will not be able to to modify " => "Multiple plugins can not modify"

"then remove and re-add those updated plugins" => "then remove and re-add the updated plugins"

www/docs/en/dev/plugin_ref/spec.md
```diff
@@ -365,6 +365,111 @@ For windows-specific attributes:
...
+
+There is an option for those who are certain that the plugin should be installed despite the conflicts. The `--force` flag can be used with `cordova plugin add`. Force adding the plugin will revert conflicting changes of other plugins so that it can be added without issues. `--force` should be used with caution as reverting changes of other plugins may cause the application to not work as expected.
+
+If the plugins ever get in a weird state, remove all plugins and re-add them.
+
+Example:
```
riknoll: Maybe make this a sub-header?

www/docs/en/dev/plugin_ref/spec.md
```diff
@@ -365,6 +365,111 @@ For windows-specific attributes:
...
+    ...
+        <activity android:configChanges="orientation|keyboardHidden" android:label="NewLabel" android:name="MainActivity">
+            ...
+        </activity>
+    ...
+    <uses-sdk android:minSdkVersion="15" android:targetSdkVersion="23" />
+</manifest>
+```
+
+Note: Reverted changes from `--force` are gone for good. They will not reappear after removing the plugin that was force added. If the reverted changes are needed, the associated plugins should be removed and re-added.
```
riknoll: "the associated plugins" => "all associated plugins"

riknoll:
That make sense to me. I added a few more comments, but otherwise LGTM.

ktop:
@riknoll I've made the changes here and will open a PR in the CLI doc to mention the plugin conflicts and link to the plugin.xml doc for reference.

riknoll:
@macdonst any chance you can take over merging this? Been sick and swamped this week and I don't think I'll have a chance to test out the site before merging it in.

PR #615, opened 2016-06-23
CB-11445 added ios release blog post

stevengill:


shazron:
We should add a note that doing a "platform update" will not reinstall your plugins if you didn't do a "--save" when you added them, or something to that effect.

stevengill:
@shazron, how does that look?

shazron:
LGTM 👍

PR #628, opened 2016-08-19
Added info to enable remote debugging in Android

PierBover:
When going to inspect devices in Chrome I could see my device connected via USB, but the Cordova webview wasn't appearing. I had to add `android:debuggable="true"` to my manifest. It took me a few hours to figure this out.


stevengill:
Thanks!

PR #634, opened 2016-09-07
Update storage.md

GillesC:
Addition of the NativeStorage plugin:
- Eliminates problems that occur when using localstorage


stevengill:
Hey @GillesC, we don't usually feature 3rd party plugins in our docs. Sorry.

Appreciate you taking the time to send a PR and maintaining such a useful plugin.

purplecabbage:
Also, the file you modified is build generated from crowdin.
Plugin documentation lives with the plugins.

PR #635, opened 2016-09-15
CB-11851 - (ios 4.3.0) Docs for developmentTeam and packageType key in build.json

shazron:
**Do not pull in** until cordova-ios 4.3.0 release.


Bwade:
@shazron Any idea when this will be merged?

PR #636, opened 2016-09-19
Update the plugman test for android

carlofragni:
The specified www directory is valid only for IOS projects, for Android projects it should target the Android source code directory.


stevengill:
Hey @carlofragni!

Plugman workflow usually means creating your application from the platforms scripts and not from the cordova-cli. For the cordova-cli, you should just use `cordova plugin add` to test plugins.

Appreciate you spending time to send a PR!

PR #642, opened 2016-09-26
CB-10929 added OXS filter button to plugin search and removed fire os…

audreyso:
CB-10929 added OXS filter button to plugin search and removed fire os filter button


dpogue:
With 10.12, Apple rebranded their desktop OS to `macOS`. Maybe we want to say `OSX / macOS`?

audreyso:
Sure! I can do that… that makes sense. Thanks for pointing that out.

stevengill:
Send a new pr. I already closed this one :)

PR #655, opened 2016-10-24
CB-12054 - Cordova iOS 4.3.0 release announcement

shazron:
cordova ios 4 3 0 apache cordova


dpogue:
Should we add a note that this requires Xcode >=7 now?

shazron:
@dpogue Good idea, I'll add it and update.

kerrishotts:
LGTM :)

PR #688, opened 2017-03-24
Remove misleading reference to "publisherId" cause it is ignored

purplecabbage:
The only way to override the `Publisher` value in the generated `AppxManifest.xml` is to provide a custom signing certificate. Merely setting the "publisherId" does not do anything. I've tried it a whole day. Only when I set a custom signing certificate my applications manifest looked like it should. And then setting "publisherId" to anything did not affect how the resulting manifest file looked.
So I propose to remove the value from the docs (and it should also be removed from the Cordova Windows plugin).

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/www/docs/en/5.1.1/guide/platforms/win8/packaging.md
+++ b/www/docs/en/5.1.1/guide/platforms/win8/packaging.md
@@ -35,7 +35,7 @@ In Windows project, identity details are kept in a file named package.appxmanife
 - Publisher
 - Version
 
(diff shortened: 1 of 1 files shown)


purplecabbage:
This pr was a replacement for #402 which I closed prematurely.

PR #691, opened 2017-03-31
CB-12609: added android@6.2.0 release blog post

stevengill:
Please review! 

I think the summary about us updating the way we use gradle could use more info. Any suggestions @infil00p?

@dpogue how does the resource-file section look?

### DIFF
--- a/www/_posts/2017-03-31-android-release.md
+++ b/www/_posts/2017-03-31-android-release.md
@@ -1,0 +1,38 @@
+---
+layout: post
+author:
+    name: Steve Gill
+    url: https://twitter.com/stevesgill
+title:  "Cordova Android 6.2.0 Released!"
+categories: announcements
+tags: news releases
+---
+
+We are happy to announce that `Cordova Android 6.2.0` has been released! 
+
+This release has fixed issues introduced by the **Android SDK Tools 25.3.1** release. **Google** dropped support for the `android` binary, so `cordova-android` has now adopted support for the `avdmanager` and `sdkmanager` binaries. We have also taken the opportunity to rewrite how we use gradle on the users system.
+
+This release also adds support for the `<resource-file>` element in `config.xml` which copies specified files during a `cordova prepare`. This allows providing arbitrary files such as special notification-sized icons, or API configuration JSON files. 
+
+To upgrade:
+
+    npm install -g cordova
+    cd my_project
+    cordova platform update android@6.2.0
+
+To add it explicitly:
+
+    cordova platform add android@6.2.0
+
+This release will have to be explicitly added until the upcoming `cordova@7` release, where it will be pinned as the default android platform.
+
+<!--more-->
+## What's new in Android
+* [CB-12614](https://issues.apache.org/jira/browse/CB-12614) Adding headers to tests
+* [CB-8978](https://issues.apache.org/jira/browse/CB-8978) Prepare copy `resource-files` from `config.xml`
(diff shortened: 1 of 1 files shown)


www/_posts/2017-03-31-android-release.md
```diff
@@ -0,0 +1,38 @@
...
+    name: Steve Gill
+    url: https://twitter.com/stevesgill
+title:  "Cordova Android 6.2.0 Released!"
+categories: announcements
+tags: news releases
+---
+
+We are happy to announce that `Cordova Android 6.2.0` has been released! 
+
+This release has fixed issues introduced by the **Android SDK Tools 25.3.1** release. **Google** dropped support for the `android` binary, so `cordova-android` has now adopted support for the `avdmanager` and `sdkmanager` binaries. We have also taken the opportunity to rewrite how we use gradle on the users system.
```
filmaj: "users" -> "user's"

www/_posts/2017-03-31-android-release.md
```diff
@@ -0,0 +1,38 @@
...
+
+This release will have to be explicitly added until the upcoming `cordova@7` release, where it will be pinned as the default android platform.
+
+<!--more-->
+## What's new in Android
+* [CB-12614](https://issues.apache.org/jira/browse/CB-12614) Adding headers to tests
+* [CB-8978](https://issues.apache.org/jira/browse/CB-8978) Prepare copy `resource-files` from `config.xml`
+* [CB-12605](https://issues.apache.org/jira/browse/CB-12605) Fix a requirements check failure on **Windows**
+* [CB-12595](https://issues.apache.org/jira/browse/CB-12595) This should find an **Android Studio** installation and use the sweet gradle center found inside
+* [CB-12546](https://issues.apache.org/jira/browse/CB-12546) leverage `avdmanager` if `android` warns it is no longer useful, which happens in **Android SDK Tools 25.3.1**. Explicitly set the `CWD` of the spawned emulator process to workaround a recent google android sdk bug. Rename `android_sdk_version.js` to `android_sdk.js`, to better reflect its contents. Have `create.js` copy over the `android_sdk_version` batch file.
```
filmaj: I'd remove this: "Rename `android_sdk_version.js` to `android_sdk.js`, to better reflect its contents. " - not useful to end users.
I would also reword the last sentence - this was actually a bug where the batch file for `android_sdk_version` would not exist in a generated cordova-android project. I am thinking: "Actually include the `android_sdk_version.bat` file in generated cordova-android projects for our lovely Windows users."

www/_posts/2017-03-31-android-release.md
```diff
@@ -0,0 +1,38 @@
...
+* [CB-12524](https://issues.apache.org/jira/browse/CB-12524) Fix for missing gradle template error. This now fetches the template from inside of the **Android Studio** directory, and falls back to a locally installed Gradle instance
```
infil00p: Explicitly state that Gradle or Android Studio are required in the blog post.  This is a change, since we used to just require the Android SDK

www/_posts/2017-03-31-android-release.md
infil00p: We should explicitly state here that we now require either Gradle or Android Studio to be installed, since our old configuration didn't require either of these things to be installed previously.

filmaj:
@infil00p's makes a great point here:

> We should explicitly state here that we now require either Gradle or Android Studio to be installed, since our old configuration didn't require either of these things to be installed previously.

We should call this out at the top-level IMO, and not within the CB bug list. I say right alongside the note about the SDK Tools 25.3.1 changes.

PR #710, opened 2017-06-12
Added instruction to install Ruby Gems locally

shazron:
### What does this PR do?

Modified instruction to install the required Ruby Gems locally in a subfolder of the repo. Installing jekyll and some modules under OS X 10.11 and up runs into problems, [especially with System Integrity Protection (SIP)](https://stackoverflow.com/questions/31567029/how-can-i-install-jekyll-on-osx-10-11/33059347#33059347)

Ran `node_modules/.bin/gulp build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,9 @@ Verify the installation by running:
 Once Ruby and JavaScript are installed, install Ruby dependencies by running:
 
     gem install bundle
-    bundle install
+    bundle install --path ./ruby_modules
 
-On some systems, the above commands need to be prefixed with `sudo`. Similarly on Windows, the `cmd` window in which those commands are to be run might need to have been "Run as Administrator."
+This will install the required Ruby Gems locally into a subfolder called `ruby_modules` in your repo folder. On some systems, the above commands need to be prefixed with `sudo`. Similarly on Windows, the `cmd` window in which those commands are to be run might need to have been "Run as Administrator."
 
 Finally, install JavaScript dependencies by running:
 
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Verify the installation by running:
 
 Once Ruby and JavaScript are installed, install Ruby dependencies by running:
 
-    gem install bundle
+    gem install bundler
     bundle install --path ./ruby_modules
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -101,9 +101,9 @@ Verify the installation by running:
 Once Ruby and JavaScript are installed, install Ruby dependencies by running:
 
     gem install bundle
-    bundle install
+    bundle install --path ./ruby_modules
```
filmaj: this wouldn't work on Windows, due to the foreslash, right?

shazron: I believe Windows can handle any slashes. @purplecabbage can you confirm? I'll test

README.md
```diff
@@ -101,9 +101,9 @@ Verify the installation by running:
 Once Ruby and JavaScript are installed, install Ruby dependencies by running:
 
     gem install bundle
-    bundle install
```
filmaj: Funnily enough, I think this is wrong (and I know it's been in here a while). I think this needs to be `gem install bundler`.

filmaj: Wait, I'm not sure about that now... my windows VM might just be freaking out.

dpogue: yes, this should be `gem install bundler`

filmaj:
Note that on my Windows 10 VM, following the exact Ruby setup instructions in here forced me to eventually update my RubyGems to a newer version - I followed [this](http://guides.rubygems.org/ssl-certificate-update/#installing-using-update-packages) on how to do that - otherwise I was unable to install `bundler` due to SSL certificate issues. Not sure if we want to call that out or otherwise update the install instructions (perhaps by recommending downloading a newer ruby installer). Possibly worth issuing a different PR for that.

However, I can confirm that `bundle install --path ./ruby_modules` runs fine on a node-js command prompt on Windows 10.

README.md
shazron: Just tested `bundle install --path ./ruby_modules` on Windows 10. It works.

shazron:
I'll update `gem install bundle` typo in this PR. I think we are good to go now. 

@filmaj yeah let's do that in another PR. Also I found that `node_modules/.bin/gulp serve` does not work on Windows, the forward slashes need to be backward slashes.

PR #713, opened 2017-07-03
CB-12980: (iOS) Improve instruction and add detail steps for plugins in cordova project.

sudotamm:
1. Embed dynamic binaries for Cordova.framework for xcode 8.3.2
2. Add step to show how to integrate plugins in xcode which used in cordova project.

### Platforms affected
Embed cordova in ios native app documentation only.

### What does this PR do?
Improve instruction and add detail steps for plugins in cordova project.

N/A

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/ios/webview.md
+++ b/www/docs/en/dev/guide/platforms/ios/webview.md
@@ -61,6 +61,8 @@ After using either of these two methods, continue with the **"Using CDVViewContr
 
 1. Add `Carthage/Build/iOS/Cordova.framework` into your Xcode project.
 
+1. Add Cordova.framework into Embedded Binaries in your Xcode General Setting page.
+
 ## 2. Adding Cleaver to the Xcode Project (CordovaLib Sub-Project)
 
 1. Quit Xcode if it is running.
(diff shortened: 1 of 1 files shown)


www/docs/en/dev/guide/platforms/ios/webview.md
```diff
@@ -220,5 +222,16 @@ After using either of these two methods, continue with the **"Using CDVViewContr
     */
     viewController.wwwFolderName = @"myfolder";
     viewController.startPage = @"mypage.html"
-    ```
+    ```
+## Adding Plugins files used in you cordova project
```
shazron: I'm not sure I agree with this section. This assumes that they already have a Cordova app. This whole section is for adding Cordova to an existing Xcode project, without needing cordova-cli and its tools. Granted a Cordova app without plugins is useless, but there is no workflow on how to do that, from start to finish. PhoneGap itself has instructions on how to do that: http://docs.phonegap.com/tutorials/develop/1-embed-webview/ios/ but that's packaging up the core plugins into Cocoapods (with a fixed config.xml). 

A generalized plugin add solution for an Embedded Webview workflow has not been implemented yet although we have discussed it (with no satisfactory conclusion).

sudotamm: When user want to integrate cordova project which used plugins in xcode project.  They also need to add source code of these plugins into their xcode project. Otherwise plugins will not work when use cordova  CDVViewController in xcode  project.

shazron: It's just that you skipped over all the parts of how they get the plugin source in the first place. That part is missing, since your steps assume a CLI workflow (assuming a Plugins folder and such).

In a CLI workflow, they just do "cordova plugin add". In an embedded webview workflow, how do they get it in without the CLI? This is the thing we haven't solved yet. By the way, just adding plugin source does not solve the problem totally, they have to add entries in config.xml, etc. It is not a trivial problem.

shazron:
@sudotamm agreed, the upper section can stay. perhaps change this PR to omit the second section, or create a new PR with just the first section.

sudotamm:
Ok, I will close this pull request and create a new one for upper section change.

PR #716, opened 2017-08-05
[Ubuntu] fix non-working commands

TomACPace:
Some commands do not work;  will continue to develop commits as I figure out how to get this to work.
This document breaks down and ceases to work at the point of the Create a click chroot environment, as the given instructions fail on 16.04.

### Platforms affected
Ubuntu 16.04

### What does this PR do?
Fix CLI commands that do not work...

I have successfully tested the changes.

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/www/docs/en/7.x/guide/platforms/ubuntu/index.md
+++ b/www/docs/en/7.x/guide/platforms/ubuntu/index.md
@@ -68,11 +68,11 @@ Once you have a basic Ubuntu VM set up, you can install the rest of the required
 
 ### Node and NPM
 
-Ubuntu 16.04 LTS comes with the required versions of both NodeJS and NPM. To install simlpy do:
+Ubuntu 16.04 LTS comes with the required versions of both NodeJS and NPM. To install simply do:
 
 ```bash
 $ sudo apt-get install nodejs npm
-$ node -v
+$ nodejs -v
 v.4.2.6
 $ npm -v
 2.15.6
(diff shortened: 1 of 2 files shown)


TomACPace:
Further commands in page are broken (ie cordova build --device fails) but attempting to solve everything...

filmaj:
Thanks for taking a stab at this! Let me know when you're happy with it and I'll merge it in.

audreyso:
Hi! @TomACPace @filmaj Is this ready to be merged? Let me know. I can help to merge.

filmaj:
The problem described here is that on Ubuntu, node.js' main command is generally not available as `node` but rather as [`nodejs`](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions) (due to `node` on apt-get being a different program altogether).

Worth nothing that this problem could affect instructions for any of the cordova tooling requiring `node` - users on Ubuntu would have the same problem.

Perhaps this PR could be tweaked so that we tailor the documentation for the generically-applicable case of `node` being the command for node.js, but point out that on Ubuntu you might need to run it as `nodejs`. I'm not sure...

What do other projects that run on node do with respect to Ubuntu?

janpio:
I think we can still merge this PR in its current state, right @filmaj @TomACPace?

janpio:
Well, now there seems to be a conflict - and no response from OP for a year.

PR #719, opened 2017-08-10
CB-13076,CB-13068 : Removed references of cordova device motion and cordova device orientation plugins

maverickmishra:
### Platforms affected
iOS, Android and Windows

### What does this PR do?
Removes references of cordova-plugin-device-motion and the cordova-plugin-device-orientation from cordova-docs

### DIFF
--- a/www/docs/en/dev/guide/support/index.md
+++ b/www/docs/en/dev/guide/support/index.md
@@ -88,17 +88,6 @@ CLI's shorthand names.
     </tr>
 
     <tr>
-        <th><a href="../../reference/cordova-plugin-device-motion/">Accelerometer</a></th>
-        <td data-col="android"    class="y"></td>
-        <td data-col="blackberry10" class="y"></td>
-        <td data-col="ios"        class="y"></td>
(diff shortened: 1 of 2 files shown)


filmaj:
So i checked this branch out, built the docs, served via `gulp serve`, and I still see the plugin API reference in the left-hand side. So we're still missing something here...

filmaj:
Here's what I see when running your branch locally via `gulp serve`:

screen shot 2017-08-10 at 6 17 23 pm

shazron:
ping @maverickmishra

maverickmishra:
I updated the plugin references from the latest `dev` version only. When the updated `dev` version is merged into 7.x, this error should get fixed.

stevengill:
hmm, @filmaj screen shot makes it seem like he is testing on dev and not on 7.x. Maybe there are other references that still need to be removed.

shazron:
@maverickmishra is this PR still relevant/updated?

filmaj:
It is. We need the docs updated to remove references to plugins that are being integrated/removed - we have more of these kinds of changes coming down the pipeline.

PR #733, opened 2017-09-21
CB-13305: Docs can't build because of ES6 code use

devgeeks:
### Platforms affected
Docs

### What does this PR do?
Updated `linkRef` function to use `var` and ES5 functions instead of
`let` and arrow functions.

Manual as well as: `gulp build --pros`

### Checklist

### DIFF
--- a/www/static/plugins/searchbar.jsx
+++ b/www/static/plugins/searchbar.jsx
@@ -8,9 +8,9 @@ var SearchBar = createClass({
 
     // polyfill of sorts for string refs
     linkRef: function(component, name) {
-        let cache = component._linkedRefs || (component._linkedRefs = {});
+        var cache = component._linkedRefs || (component._linkedRefs = {});
         if (!component.refs) component.refs = {};
-        return cache[name] || (cache[name] = c => {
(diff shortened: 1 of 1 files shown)


www/static/plugins/searchbar.jsx
```diff
@@ -8,9 +8,9 @@ var SearchBar = createClass({
 
     // polyfill of sorts for string refs
     linkRef: function(component, name) {
-        let cache = component._linkedRefs || (component._linkedRefs = {});
+        var cache = component._linkedRefs || (component._linkedRefs = {});
         if (!component.refs) component.refs = {};
-        return cache[name] || (cache[name] = c => {
+        return cache[name] || (cache[name] = function(c) {
             component.refs[name] = c;
```
shazron: should this be:
`return component.refs[name] = c;` ? if not the assignment doesn't make sense

shazron: Or is the cache caching a function? (I'm going by what arrow functions typically do)

devgeeks: Ah hah. Good catch. Arrow functions have implicit returns.

PR #737, opened 2017-09-26
CB-13333: Move "Instructions" to /doc

janpio:
Move instructions about blogposts, redirects and adding tools or showcase apps to individial files in /doc

(Only minimal changes to actual instructions)

### DIFF
--- a/doc/redirects.md
+++ b/doc/redirects.md
@@ -1,0 +1,53 @@
+# Redirects
+
+Sometimes (docs) pages get removed, renamed, and added. There is [a redirects file](../www/_data/redirects.yml) that contains redirects for such occasions. It has three sections:
+
+- `general`: single-page redirects,
+- `docs-global`: redirects across all docs versions and languages, and
+- `docs`: version-specific docs redirects.
+
+For non-docs URIs, there are no versioning considerations. Make redirects like so:
+
+    general:
+        - {old: "old/uri/for/page.html", new: "its/new/uri.html"}
+
+**NOTE**: Review (and test, if possible) these redirects before making them live, since they're permanent (HTTP 301) redirects. Incorrect permanent redirects will make a URI almost impossible to bring back into browsers and search indices.
+
+## Changing Docs URIs
+
+There are five cases of changing URIs:
+
+1. Adding a brand new (no past equivalent) URI starting at a version,
+2. Removing an old URI (with no replacement) starting at a version,
+3. Renaming (removing and adding) an existing URI starting at a version,
+4. Renaming an existing URI across all versions,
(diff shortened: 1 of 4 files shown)


doc/redirects.md
```diff
@@ -0,0 +1,53 @@
+# Redirects
+
+Sometimes (docs) pages get removed, renamed, and added. There is [a redirects file](../www/_data/redirects.yml) that contains redirects for such occasions. It has three sections:
+
+- `general`: single-page redirects.
+- `docs-global`: redirects across all docs versions and languages,
+- `docs`: version-specific docs redirects, and
```
dblotsky: The items in this list are in the wrong order.

doc/redirects.md
```diff
@@ -0,0 +1,53 @@
...
+
+For non-docs URIs, there are no versioning considerations. Make redirects like so:
+
+    general:
+        - {old: "old/uri/for/page.html", new: "its/new/uri.html"}
+
+**NOTE**: Review (and test, if possible) these redirects before making them live, since they're permanent (HTTP 301) redirects. Incorrect permanent redirects will make a URI almost impossible to bring back into browsers and search indices.
+
+## Changing Docs URIs
```
dblotsky: Why is this a new section?

doc/redirects.md
janpio: Hm, strange how this happened, these are indeed different than in the original.
But why is this "wrong", what defines the order?

doc/redirects.md
janpio: I felt the list of cases below required a "parent" headline. Is it factually wrong? Or should the `general` above also get one?

janpio:
I hope not - having clearly defined files should make it easier to read the stuff that actually matters. Of course we might want to link to it (and highlight its importance) at the correct places. 

(The README in general needs additional work for that. Maybe a list of "Guides" or "Common Tasks" after the introduction and before the install/build/develop/deploy [which maybe also could be extracted to their own files?)

doc/redirects.md
dblotsky: Just the commas and the "and". I'm just pointing out that the writing is currently wrong, so it just needs to be fixed to match, or the order should be back to what it was before. The actual order doesn't matter.

doc/redirects.md
dblotsky: It's not wrong. It was just all one section before, so I'm wondering why you split it into two sections. Both are OK.

doc/redirects.md
janpio: Oh, right. Fixed.

PR #740, opened 2017-09-28
CB-13351: resort and group attributes of widget table

janpio:
The attributes had no recognizable order. I resorted and grouped them so all the "build number" and "package name" attributes are together.


janpio:
(Unfortunately Markdown tables don't support any separators, so the grouping can not easily be visualized)

PR #746, opened 2017-10-05
CB-13054: blog post on how to move off of file-transfer plugin

filmaj:
FYI / please review @audreyso, @macdonst, @stevengill.

### DIFF
--- a/www/_posts/2017-10-18-from-filetransfer-to-xhr2.md
+++ b/www/_posts/2017-10-18-from-filetransfer-to-xhr2.md
@@ -1,0 +1,178 @@
+---
+layout: post
+author:
+    name: Fil Maj
+    url: https://twitter.com/filmaj
+title:  "Transition off of cordova-plugin-file-transfer"
+categories: blog
+tags: plugins
+---
+
+Early on in Cordova's existence, the [file-transfer plugin](https://github.com/apache/cordova-plugin-file-transfer)
+was created to solve the [problem of downloading binary files](https://issues.apache.org/jira/browse/CB-22).
+At the time, there weren't great options for solving this using standards-compliant
+web APIs. The web took a twisty path to get to a solution (see
+[Firefox's `sendAsBinary`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/sendAsBinary)
+and the now-defunct [FileSystem API](https://dev.w3.org/2009/dap/file-system/file-dir-sys.html)'s
+[BlobBuilder](https://developer.mozilla.org/en-US/docs/Web/API/BlobBuilder),
+among others), but today you can use our good friend [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)'s
+newest features, combined with some newer JavaScript types and objects,
+[to solve this problem](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data).
+This is an exciting moment for Cordova as the dream for this project was always
+to eventually reduce the surface area of APIs the project maintains, and instead
+see regular web APIs be able to handle these use cases.
+
+As a result,
+[Cordova is sunsetting the file-transfer plugin](https://issues.apache.org/jira/browse/CB-13052).
+What does "sunsetting" mean? In summary:
+
+* No more work will be done on the file-transfer plugin by the Cordova development
+  community.
+* You can continue to use the file-transfer plugin if you wish - it should work
+  fine as-is for the foreseeable future.
+* We highly suggest Cordova users transition to using the
+  [standards-compliant way of sending and receiving binary data](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data).
+
+All of us at Apache Cordova don't want to leave y'all hanging, though, so we
+thought it'd be a good idea to show you how to use these newer XHR features to do
+what file-transfer lets you do, but in a way that will work in any modern web
+browser to boot!
+
+## Requirements
+
+Based on how deeply you interact with the underlying device filesystem, and on
+which platforms, you may still need to rely on the
+[Cordova File plugin](https://github.com/apache/cordova-plugin-file). If you
+still have references to `requestFileSystem` or `root.fs` in your application's
+JavaScript, you will definitely need the File plugin because these are not
+standards-compliant APIs. Take note and care!
+
+## Platform Support
+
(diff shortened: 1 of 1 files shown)


www/_posts/2017-10-02-from-filetransfer-to-xhr2.md
```diff
@@ -0,0 +1,85 @@
...
+  * It [uses](https://github.com/filmaj/xhr2test/blob/master/www/js/index.js#L31)
+    [localforage](https://github.com/localForage/localForage) as a storage abstraction.
+  * The [`helpers.js` file](https://github.com/filmaj/xhr2test/blob/master/www/js/helpers.js)
+    has two methods you'll need:
+    * [`download_blob_via_xhr`](https://github.com/filmaj/xhr2test/blob/master/www/js/helpers.js#L59)
+    * [`upload_blob_via_xhr`](https://github.com/filmaj/xhr2test/blob/master/www/js/helpers.js#L42)
+
+## The Long Version
+
+If you want to understand some of the nuts and bolts enabling binary data transfering,
```
audreyso: transferring* :)

macdonst:
@filmaj looks good to me other than the file name. Does `2017-10-02` mean it will show up in the past on the blog? Maybe update that to the day you publish?

janpio:
I (unfairly) expected it to be a post with "replace `upload` with this code, replace `download` with this code", maybe even an API identical implementation of the new stuff. Is your [`helpers.js` file](https://github.com/filmaj/xhr2test/blob/master/www/js/helpers.js) something like this?

macdonst:
@janpio where the plugins are not quite spec complaint we will be updating the plugins to match current standards. In the case of the FileTransfer plugin which was created out of necessity and doesn't follow any spec we are not providing backward compatibility. Even though we won't be doing anything with the plugin in the future it will still continue to work for folks. If one of the community members wants to take on a wrapper lib that would be something I'd support.

janpio:
Ok, so let's see if I got it right:
- There will be no more dev work on the plugin as there is now a standard alternative. 
- There is no pressure on switching to the standard as this all works just fine right now.
- Only in the future with new releases of iOS or Android it might break. 
- Switching from this to the standard is a bit of work _right now_, but could be simplified a lot if someone did a wrapper for it that uses the same API as the plugin.

All correct?

filmaj:
@macdonst sounds good, once this gets the thumbs up I'll change the date to make it the day-of.

@janpio yeah, the helpers.js file bits are supposed to be drop-in replacements, but only for the most common uses of the file-transfer plugin. The plugin supports many options - this would be a long post if we wanted to cover everything. That said, I'm open to adding or changing the post to address your concern. I could potentially change the "TL;DR" section to instead just be drop-in replacements for the main `upload` and `download` methods, and write XHR-based equivalents for [the main examples for these methods that exit in our docs](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-file-transfer/index.html#example).

Should we call out the fact that there will be no more work done on the file transfer plugin? I thought that was implied when I mentioned we were sunsetting it.

macdonst:
@janpio yup, I feel like that is a good summary.

janpio:
To be honest to me in practice "sunsetting" or "deprecating" never is really clear - even if I theoretically know what it means. My thoughts when reading: Does it mean I have to change my stuff now? That it's broken? That something doesn't or won't work? What does "no more maintance" (or something similar) really mean?
Now I know it just means "we stop future development as it works just fine right now, for the future a better standard alternative is here that can replace all this stuff (and is maintained by the browser directly!). so you don't need this any more, but feel free to keep using it.".

macdonst:
@janpio I think your English is great and I kinda want _"we stop future development as it works just fine right now, for the future a better standard alternative is here that can replace all this stuff (and is maintained by the browser directly!). so you don't need this any more, but feel free to keep using it."_ on a t-shirt

filmaj:
@janpio that makes sense. Us as cordova committers probably have way more context around what sunsetting means because we've been talking about it for months! I will take a stab at rewriting that part.

Also curious to hear your thoughts about the code examples. Are links to my PoC app good enough, or do we want something more fleshed out? I don't want to half-ass this :)

janpio:
Thanks guys ;)

Thing is if you do more than your PoC app that covers the standard usecase (like e.g. putting it as a library on Github, installable via npm, which people can use as a drop in replacement, fork and contribute their own more complicated use cases) you have to maintain that one, accept PRs etc - which kind of misses the point of sunsetting the plugin ;)

But having the TLDR more accessible like you suggested in https://github.com/apache/cordova-docs/pull/746#issuecomment-334767623 sure would go a long way.

Many people use this plugin via the Ionic Native wrapper in Ionic/Angular, [no idea yet what will happen there](https://github.com/ionic-team/ionic-native/issues/2026) (as it is also community maintained).

filmaj:
I made a bunch of changes to it @macdonst @janpio. Also explicitly called out support for Windows and eliminated any links to my xhr test app as per feedback from @purplecabbage.

www/_posts/2017-10-02-from-filetransfer-to-xhr2.md
```diff
@@ -0,0 +1,175 @@
...
+* Windows Phone 8 or newer.
+
+As always, check [caniuse.com](https://caniuse.com) for detailed support for the
+required bits, like [`Blob`](http://caniuse.com/#feat=blobbuilder),
+[Typed Arrays](http://caniuse.com/#feat=typedarrays), and
+[extended XHR features](http://caniuse.com/#feat=xhr2).
+
+## TL;DR
+
+Don't care about "standards" and all that mumbo jumbo? Just looking to copy-paste
```
janpio: Sounds a bit snarky. Intended?

janpio:
I like it.

www/_posts/2017-10-02-from-filetransfer-to-xhr2.md
filmaj: Sort of.. suggestions on how to reword?

janpio: "Standards are great and all, but what do you actually have to copy-paste to replace the previous FileTransfer examples? We got you covered:

Here's ..."

filmaj:
@janpio updated based on your suggestion - thank you for that.

@macdonst I know you are very busy this week, but pinging you here to put this on your radar. Please don't rush to get to this one, at your leisure, but, once I get your thumbs-up, I'll change the filename to reflect the current date and merge it in.

✌️

macdonst:
@filmaj sorry, this got buried last week while I was away. The blog post looks great. I'd say release it.

PR #748, opened 2017-10-13
[WIP] CB-13440: Improve "Using Flags" for iOS

janpio:
- move "Provisioning Profile" down as it is not to be used any more with current xcode
- move "Development Team" and "Packaging Type" above optional value
- multiple line breaks in descriptions
- small text changes in descriptions

Drive by changes:
- Clarify "Using build.json": Also for newer versions of Xcode and iOS
- Reword "Xcode Build Flags" a bit and link to "Using build.json" headline instead of just mentioning it


janpio:
Hmm, I am not sure any more if my "pre Xcode8" vs "Xcode 8+" is correct for the flags. Should check again.

janpio:
(And now me being an idiot I even deleted the forked repo so I can't update the PR... will have to create a new one when I understood what is actually going on)

PR #765, opened 2017-11-16
ShowCase: Remove broken app, add SparkChess

armandn:
Added SparkChess to the Showcase. SparkChess is an an app with over 100K users available for both iOS and Android, built with Cordova. (it's also available online). Thank you.


jcesarmobile:
@shazron I'm personally +1, but I think we decided we weren't adding more apps anymore

armandn:
Well, if it helps, the ReactEurope app doesn't work (http://www.thinkmill.com.au/react-europe results in 404), so you could replace it...

janpio:
@armandn Could you please open another PR for that change as well? Then I think we can go ahead and merge this.

armandn:
I removed the non-working ReactEurope and committed but the commit was added to this PR... I'm no expert in Github, not sure if I've done something wrong...

janpio:
You forked the repo and made your changes to master, then opened a pull request for your master. Now you added another commit to that master, which changed this pull request.

But it's fine I guess.

janpio:
Updated the PR to title to match the new reality.

jcesarmobile:
We should probably do a cleanup, the other day I saw on twitter that Gudog was a React Native app now

janpio:
removing the whole section maybe?

armandn:
Actually having a showcase helps when evaluating a technology, framework or whatever. If anything, I'd make it its own section, like Electron has: https://electronjs.org/apps . Also Three.js .

PR #775, opened 2017-12-04
CB-13620: 7.0.0 Docs Update.  Includes upgrade instructions.

infil00p:
### Platforms affected
Android

### What does this PR do?
This adds the announcement for the 7.0.0 release

### Checklist
- [] Added automated test coverage as appropriate for this change.

### DIFF
--- a/www/_posts/2017-12-04-cordova-android-7.0.0.md
+++ b/www/_posts/2017-12-04-cordova-android-7.0.0.md
@@ -1,0 +1,58 @@
+---
+layout: post
+author:
+    name: Joe Bowser
+    url: https://twitter.com/infil00p
+title:  "Cordova Android 7.0.0 Released!"
+categories: announcements
+tags: news releases
+---
+
+We are happy to announce that `Cordova Android 7.0.0` has been released! 
+
+With this release, we have changed the default project structure for Android projects.  People who currently use the CLI and treat everything in the platforms directory as a 
+build artifact should not notice a difference, however this a major breaking change for people creating standalone Cordova Android projects.  This also means that the locations
+of files have changed and have been brought in line to the structure used by Android Studio.
+
+## Major Changes include:
+ * Support for Java 1.8 language features in Cordova Plugins
+ * CordovaInterface now has a Context getter so that contexts can be retrieved without an Activity
+ * Cordova can now build for x86_64, arm64 and armeabi architecture when building plugins that use the NDK
+ * The minimum Android API version supported is now API Level 19
+ * Due to the directory structure change, we no longer support in-line upgrading, bringing us in line with iOS
+ * ANT builds are no longer supported and the functionality has been removed.
+
+To upgrade:
+
+    npm install -g cordova
+    cd my_project
+    cordova platform remove android
+    cordova platform add android@7.0.0
+
+To add it explicitly:
+
+    cordova platform add android@7.0.0
+
+<!--more-->
+## Curated Changelog
+* [CB-13612](https://issues.apache.org/jira/browse/CB-13612) Fix the remapper so that XML files copy over and the Camera works again.
+* [CB-13741](https://issues.apache.org/jira/browse/CB-13741) Bump `package.json` so we can install plugins
+* [CB-13610](https://issues.apache.org/jira/browse/CB-13610) Compress the default app assets
+* [CB-12835](https://issues.apache.org/jira/browse/CB-12835) add a Context getter in CordovaInterface
+* [CB-8976](https://issues.apache.org/jira/browse/CB-8976) Added the `cdvVersionCodeForceAbiDigit` flag to the template build.gradle that appends 0 to the versionCode when `cdvBuildMultipleApks` is not set
+* [CB-12291](https://issues.apache.org/jira/browse/CB-12291) (android) Add x86_64, arm64 and armeabi architecture flavors
+* [CB-13602](https://issues.apache.org/jira/browse/CB-13602) We were setting the path wrong, this is hacky but it works
+* [CB-13601](https://issues.apache.org/jira/browse/CB-13601) Fixing the standalone run scripts to make sure this works without using the CLI
+* [CB-13580](https://issues.apache.org/jira/browse/CB-13580) fix build for multiple apks (different product flavors)
+* [CB-13558](https://issues.apache.org/jira/browse/CB-13558) Upgrading the gradle so we can upload the AAR
+* [CB-13297](https://issues.apache.org/jira/browse/CB-13297) This just works once you bump the project structure.  Java 1.8 compatibility baked-in
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) **Android** Studio 3 work, things have changed with how the platform is built
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Found bug where the gradle subproject changes weren't actually getting written to the correct gradle file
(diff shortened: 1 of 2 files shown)


www/docs/en/dev/guide/platforms/android/upgrade.md
```diff
@@ -27,6 +27,48 @@ Most of these instructions apply to projects created with an older set
 of command-line tools that precede the `cordova` CLI utility. See [The Command-Line Interface](../../cli/index.html) for information how to update the
 version of the CLI.
 
+
+## Upgrading to 7.X.X
+
+The best way to upgrade to 5.X.X is to simply remove the Android platform from
```
janpio: 5.X.X => 7.X.X

www/docs/en/dev/guide/platforms/android/upgrade.md
```diff
@@ -27,6 +27,48 @@ Most of these instructions apply to projects created with an older set
...
+```
+
+If you use the above method, be aware that any changes you made to the android
+platform folder will be lost (editing the contents of this folder is
+discouraged).
+
+Alternatively, you may attempt to use the platform update script. For non-CLI
+projects, run:
+
+``
```
janpio: Not enough \`
(same 2 lines below)

www/_posts/2017-12-04-cordova-android-7.0.0.md
```diff
@@ -0,0 +1,58 @@
...
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) **Android** Studio 3 work, things have changed with how the platform is built
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Found bug where the gradle subproject changes weren't actually getting written to the correct gradle file
+* [CB-13470](https://issues.apache.org/jira/browse/CB-13470) Fix Clean so that it cleans the **Android** Studio structure
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Adding specs for resource files inside an **Android** Studio Project
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Added remapping for drawables
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Found bug in Api.js where xml/strings.xml is used instead of values/strings.xml
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Setup Api.js to support multiple builders based on project structure
+* [CB-11244](https://issues.apache.org/jira/browse/CB-11244) Changing directory creation, will most likely hide this behind a flag for the next release of `cordova-android`, and then make it default in the next major pending feedback
```
janpio: Is everything starting with "will most ..." expected to be here?

www/_posts/2017-12-04-cordova-android-7.0.0.md
```diff
@@ -0,0 +1,58 @@
...
+* Adding the Studio Builder to build a project based on **Android** Studio, and deleting Ant, since Google does not support Ant Builds anymore. Sorry guys!
```
janpio: Sorry guys => Sorry folks

www/_posts/2017-12-04-cordova-android-7.0.0.md
```diff
@@ -0,0 +1,58 @@
...
+title:  "Cordova Android 7.0.0 Released!"
+categories: announcements
+tags: news releases
+---
+
+We are happy to announce that `Cordova Android 7.0.0` has been released! 
+
+With this release, we have changed the default project structure for Android projects.  People who currently use the CLI and treat everything in the platforms directory as a 
+build artifact should not notice a difference, however this a major breaking change for people creating standalone Cordova Android projects.  This also means that the locations
+of files have changed and have been brought in line to the structure used by Android Studio.
```
dpogue: We probably need to add something here to tell people that `edit-config` tags in plugin.xml and config.xml will need their paths updated?

i.e.,
```xml
<!-- An existing config.xml -->
<edit-config file="AndroidManifest.xml" target="/manifest/application" mode="merge">

<!-- needs to change to -->
<edit-config file="app/src/main/AndroidManifest.xml" target="/manifest/application" mode="merge">
```

www/_posts/2017-12-04-cordova-android-7.0.0.md
infil00p: You are aware that this is taken directly from the commit messages, right?

www/_posts/2017-12-04-cordova-android-7.0.0.md
infil00p: Yes, it's the commit message.

www/_posts/2017-12-04-cordova-android-7.0.0.md
infil00p: That's a good idea.

www/_posts/2017-12-04-cordova-android-7.0.0.md
janpio: No, as it says "Curated Changelog" as headline above.

The commit message was similarly problematic then. No need to keep this in the blog post though.

infil00p: @janpio The commit is in the changelog because dropping ANT support is important.  I've edited it.  It was an innocent mistake that I can't fix without violating a bunch of ASF policies.

www/_posts/2017-12-04-cordova-android-7.0.0.md
dpogue: Actually, if you search GitHub for `edit-config`, it seems like 99% of them are editing AndroidManifest.xml. Is it possible to special case that so it uses the right path? Did that already happen in cordova-common and we're waiting for a tools release?

www/_posts/2017-12-04-cordova-android-7.0.0.md
janpio: Removing the word - as you did - is totally fine, thank you.

www/_posts/2017-12-04-cordova-android-7.0.0.md
infil00p: I don't believe any of edit-config's functionality has been changed.  Of course, we only test on our own plugins, and none of those use edit-config on Android, only on iOS.  I'm sure it's possible to add this special case,  but that would mean that we would be on the hook for maintaining it, when people really should be editing the path.  I think we wrote enough new workaround code for this change already.

dpogue: Then we should also mention that it's not possible for a plugin using `edit-config` to support both cordova-android@6 and cordova-android@7?

infil00p: Fair enough.  That said, what's an example plugin that is widely used that would be broken by this?

macdonst: CSDK plugins are affected by this but they are not being actively maintained.

dpogue: https://github.com/macdonst/cordova-plugin-allow-backup is one, I know I've used edit-config in a few of my company's internal plugins

macdonst: @dpogue no one needs to use cordova-plugin-allow-backup as they can just add those lines in config.xml instead.

infil00p: OK, I added a brief mention after the first paragraph and copy-pasted your code example.  BTW: We can all edit the PR in GitHub itself.

PR #791, opened 2018-02-27
CB-13924: added version 8 of docs

stevengill:


dblotsky:
@stevengill did you do a `gulp clean; gulp fetch` to make sure you have the latest version of the plugin docs?

stevengill:
@dblotsky no, but i shall

dblotsky:
@stevengill do you plan to deploy also?

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+    name: Create apps
+    children:
+        - url: guide/cli/index.html
+        - url: guide/cli/template.html
+        - url: guide/support/index.html
+        -
+            name: Develop for platforms
+            children:
+                - url: guide/platforms/android/index.html
+                - url: guide/platforms/blackberry10/home.html
```
janpio: we should probably remove the deprecated platforms here, should we?

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+                - url: guide/platforms/ios/index.html
+                - url: guide/platforms/osx/index.html
+                - url: guide/platforms/ubuntu/index.html
+                - url: guide/platforms/win8/index.html
+                - url: guide/platforms/wp8/home.html
```
janpio: see above (deprecated platform)

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+        - url: guide/cli/template.html
+        - url: guide/support/index.html
+        -
+            name: Develop for platforms
+            children:
+                - url: guide/platforms/android/index.html
+                - url: guide/platforms/blackberry10/home.html
+                - url: guide/platforms/ios/index.html
+                - url: guide/platforms/osx/index.html
+                - url: guide/platforms/ubuntu/index.html
```
janpio: see above (deprecated platform)

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+                - url: guide/platforms/win8/index.html
```
janpio: can we change this URL to `windows` instead of `win8` somehow?

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+        - url: guide/appdev/whitelist/index.html
+-
+    name: Create plugins
+    children:
+        - url: guide/hybrid/plugins/index.html
+        -
+            name: Develop for platforms
+            children:
+                - url: guide/platforms/android/plugin.html
+                - url: guide/platforms/blackberry10/plugin.html
```
janpio: see above (deprecated platform)

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+                - url: guide/platforms/ios/plugin.html
+                - url: guide/platforms/win8/plugin.html
+                - url: guide/platforms/wp8/plugin.html
```
janpio: see above (deprecated platform)

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+    name: Create plugins
+    children:
+        - url: guide/hybrid/plugins/index.html
+        -
+            name: Develop for platforms
+            children:
+                - url: guide/platforms/android/plugin.html
+                - url: guide/platforms/blackberry10/plugin.html
+                - url: guide/platforms/ios/plugin.html
+                - url: guide/platforms/win8/plugin.html
```
janpio: see above (rename to `windows`)

www/_data/toc/en_8-x-src.yml
dblotsky: If you deprecate it, make sure to add it to `redirects.yml`.

www/_data/toc/en_8-x-src.yml
dblotsky: @janpio we can, but make sure to redirect in `redirects.yml`

www/_data/toc/en_8-x-src.yml
```diff
@@ -0,0 +1,76 @@
...
+    name: Reference
+    children:
+        - url: config_ref/index.html
+        - url: cordova/events/events.html
+        - url: reference/cordova-cli/index.html
+        - url: guide/appdev/hooks/index.html
+        - url: plugin_ref/spec.html
+        -
+            name: Plugin APIs
+            children:
```
janpio: we should probably remove the deprecated plugins here, should we?

stevengill:
@dblotsky i was planning on letting travis do the deploy

dblotsky:
@stevengill then that fetching issue might still be ok, if Travis does a clean build every time. However even for local development, I've made a fix so that you don't have to remember to keep doing this: https://github.com/apache/cordova-docs/pull/793.

stevengill:
@janpio I agree we should remove the deprecated platforms and plugins. I'm just struggling figuring out where the removal should happen. The file you commented on is generated. So I don't think removing it from this file is the way to go. It would be better if we removed it from the generation step

stevengill:
I'm also not sure how removing those fields will affect past versions of the docs. Maybe @dblotsky can chime in

stevengill:
I'm trying to remove the fields from https://github.com/apache/cordova-docs/blob/master/www/_data/fetched-files.yml.

janpio:
Yeah, I thought so as well but couldn't figure out in a few minutes how to do this - so I commented on the first appearance of the "wrong" files/links.

dblotsky:
@janpio @stevengill that's the correct file: https://github.com/apache/cordova-docs/blob/master/www/_data/fetched-files.yml.

When you run `gulp fetch`, the files listed in that file get downloaded to `dev/en/{file dest}`. You can see more of this by running `node ./tools/bin/fetch_docs.js`.

Maybe we can put together a flowchart of the build process as it is right now. Would that help you guys?

dblotsky:
I kind of assumed that people would run `gulp` and `gulp --tasks`, or read `gulpfile.js` to see what controls what in the build. However, I agree that a text-based or graphical explanation might be useful as well.

stevengill:
Closing this PR and opening a new one for this

PR #799, opened 2018-03-02
fix some internal links

janpio:
I fixed some internal links that were broken in the dev version of the docs.


dblotsky:
@janpio have you manually tested all of these links? If so, looks good to me.

janpio:
@dblotsky Yes, this is from a local crawl of http://localhost:3000/docs/en/dev/, then fixing broken links that seemed fixable.

PR #800, opened 2018-03-02
document jekyll configuration in `conf` 

janpio:
README that makes it easier to discover what `conf` is and the files mean.


conf/README.md
```diff
@@ -0,0 +1,17 @@
...
+- `_version.yml` defines the latest version of the docs
+- `_dev.yml` is used for normal builds
+- `_prod.yml` is used when the `--prod` option is used for the build
+- `_nodocs.yml` is used when the `--nodocs` option is used for the build
+
+Some files are generated by the build process itself:
+
+- `_defaults.yml` defines different config values for different parts of the docs
+
+To learn more about how to control the build with these files, have a look at: https://jekyllrb.com/docs/configuration/
```
dblotsky: I'd vote for putting this line in the top paragraph, and making it bold. :)

conf/README.md
```diff
@@ -0,0 +1,17 @@
+# Jekyll Configuration
+
+This folder contains different configuration files for the Jekyll build, which is triggered via `gulp build` and similar commands (`gulp serve` etc.).
+
+Files present after checkout:
+
+- `_config.yml` is the main Jekyll config file
+- `_version.yml` defines the latest version of the docs
+- `_dev.yml` is used for normal builds
```
dblotsky: It might be better to say it's used for build by *default*, because "normal" builds are actually production builds, not dev builds. Dev builds are special.

Also, `_dev.yml` and `_prod.yml` are mutually exclusive: when `--prod` is specified, `_dev.yml` is not included.

janpio:
Well, I would try to update it if changes to the files happen - I can't speak for other devs working on the project. I just know it would have helped me understand the project structure and ways to configure a lot if I had known that these are Jekyll config files and what they do.

PR #801, opened 2018-03-02
quicker dev build

janpio:
use exclude/include to only build 3 languages in dev 
instead of using exclusion list that always has to be updated

### DIFF
--- a/conf/_dev.yml
+++ b/conf/_dev.yml
@@ -1,107 +1,13 @@
 baseurl: ""
 default_linked_docs_version: dev
 destination: build-dev
+
 exclude:
     - static/css-src
     - static/plugins
-    - docs/de/3.1.0
-    - docs/de/3.4.0
-    - docs/de/3.5.0
-    - docs/de/5.4.0
-    - docs/de/6.x
-    - docs/de/dev
-    - docs/en/1.5.0
-    - docs/en/1.6.0
-    - docs/en/1.6.1
-    - docs/en/1.7.0
-    - docs/en/1.8.0
-    - docs/en/1.8.1
-    - docs/en/1.9.0
-    - docs/en/2.0.0
-    - docs/en/2.1.0
-    - docs/en/2.2.0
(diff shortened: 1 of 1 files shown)


conf/_dev.yml
```diff
@@ -1,107 +1,13 @@
...
-    - docs/zh-tw/3.5.0
-    - docs/zh-tw/5.4.0
-    - docs/zh-tw/6.x
-    - docs/zh-tw/dev
+    - docs/
+
+include:
+    - docs/en/dev
+    - docs/ru/dev
+    - docs/fr/dev
```
dblotsky: Nitpick: no newline at the end of the file.

conf/_dev.yml
```diff
@@ -1,107 +1,13 @@
...
-    - docs/zh-cn/dev
-    - docs/zh-tw/3.1.0
-    - docs/zh-tw/3.4.0
-    - docs/zh-tw/3.5.0
-    - docs/zh-tw/5.4.0
-    - docs/zh-tw/6.x
-    - docs/zh-tw/dev
+    - docs/
+
+include:
```
dblotsky: It might also be worth it to add the `7.x` or `6.x` versions, so that it's easier to test some links.

janpio: Would having those for `en` be enough?

dblotsky: @janpio so, I'd prefer these:
```
en/dev
ru/dev
fr/dev
en/7.x
ru/7.x
fr/7.x
```

Menardi:
I was just about to create a PR for this. I wish this had already been merged, it would have saved me a lot of time! Is there are reason it hasn't been merged yet?

janpio:
See the second comment. Also I am not sure if the release of the 8.x docs has changed anything that would have to be adapted here. Feel free to drive this forward (as I am not free right now)!

dblotsky:
@janpio this looks good to merge, even without the other versions.

PR #806, opened 2018-03-15
CB-13964: Fixed make clean for windows

gandhirajan:
This PR fixes 'make clean' command issue in windows OS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,9 @@ endif
 ifdef WINDOWS
 RM = cmd /C del /Q /F $(subst /,\,$(1))
 RMDIR  = cmd /C rmdir /S /Q $(subst /,\,$(1))
+else
+RM = rm -f $(1)
+RMDIR = rm -rf $(1)
 endif
 
 # constants
@@ -292,7 +295,6 @@ $(CSS_DEST_DIR)/%.css: $(CSS_SRC_DIR)/%.css
 	$(call printfile,$<) >> $@
 
 # maintenance
-ifdef WINDOWS
 clean:
 	$(call RM, $(VERSION_CONFIG))
 	$(call RM, $(DEFAULTS_CONFIG))
@@ -304,18 +306,6 @@ clean:
 	-$(call RMDIR, $(CSS_DEST_DIR))
 	-$(call RMDIR, $(PROD_DIR))
 	-$(call RMDIR, $(DEV_DIR))
-else
-clean:
-	$(RM) $(VERSION_CONFIG)
-	$(RM) $(DEFAULTS_CONFIG)
-	$(RM) $(DOCS_PAGE_LIST)
-	$(RM) $(DOCS_VERSION_DATA)
-	$(RM) -r $(PROD_DIR) $(DEV_DIR)
-	$(RM) $(TOC_FILES)
-	$(RM) $(PLUGINS_APP)
-	$(RM) -r $(CSS_DEST_DIR)
-	$(RM) $(FETCHED_FILES)
-endif
 
 nuke: clean
 	$(RM) -r node_modules
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ SHELL  = cmd
 JEKYLL = bundle.bat exec jekyll
 CAT    = type
 LS     = ls
+RM = cmd /C del /Q /F
+RMDIR = cmd /C rmdir /S /Q
 else
 SHELL  = sh
 JEKYLL = bundle exec jekyll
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,6 +14,8 @@ SHELL  = cmd
 JEKYLL = bundle.bat exec jekyll
 CAT    = type
 LS     = ls
+RM = cmd /C del /Q /F
+RMDIR = cmd /C rmdir /S /Q
```
dblotsky: Nitpick: align equals signs.

Makefile
```diff
@@ -154,6 +156,17 @@ JEKYLL_FLAGS   =
...
+ifdef WINDOWS
+	VERSION_CONFIG       := $(subst /,\,$(VERSION_CONFIG))
+	DEFAULTS_CONFIG      := $(subst /,\,$(DEFAULTS_CONFIG))
+	DOCS_PAGE_LIST       := $(subst /,\,$(DOCS_PAGE_LIST))
+	DOCS_VERSION_DATA    := $(subst /,\,$(DOCS_VERSION_DATA))
+	TOC_FILES            := $(subst /,\,$(TOC_FILES))
+	PLUGINS_APP          := $(subst /,\,$(PLUGINS_APP))
+	CSS_DEST_DIR         := $(subst /,\,$(CSS_DEST_DIR))
+	FETCHED_FILES        := $(subst /,\,$(FETCHED_FILES))
+endif
```
dblotsky: This could break other uses of these files. Instead, change the definition of `RM` and `RMDIR` to escape slashes, like how `makedir` does it [here](https://github.com/apache/cordova-docs/blob/master/Makefile#L36).

Makefile
```diff
@@ -287,6 +300,19 @@ $(CSS_DEST_DIR)/%.css: $(CSS_SRC_DIR)/%.css
...
+	$(RM) $(VERSION_CONFIG)
+	$(RM) $(DEFAULTS_CONFIG)
+	$(RM) $(DOCS_PAGE_LIST)
+	$(RM) $(DOCS_VERSION_DATA)
+	$(RM) $(TOC_FILES)
+	$(RM) $(PLUGINS_APP)
+	$(RM) $(FETCHED_FILES)
+	IF EXIST "$(CSS_DEST_DIR)" $(RMDIR) $(CSS_DEST_DIR)
+	IF EXIST "$(PROD_DIR)" $(RMDIR) $(PROD_DIR)
+	IF EXIST "$(DEV_DIR)" $(RMDIR) $(DEV_DIR)
```
dblotsky: As above, you can avoid the `if`/`else` here by defining `RM` differently. Also, you can avoid the `IF EXIST` by adding a `-` to the beginning of the line, which will tell Make to ignore any errors. So:
```
-$(RMDIR) $(CSS_DEST_DIR)
```
instead of 
```
$(RMDIR) $(CSS_DEST_DIR)
```

Makefile
```diff
@@ -297,9 +323,10 @@ clean:
...
 	$(RM) -r $(CSS_DEST_DIR)
 	$(RM) $(FETCHED_FILES)
+endif
 
 nuke: clean
 	$(RM) -r node_modules
 	$(RM) Gemfile.lock
 
-.PHONY: clean usage help default build fetch $(DEV_DOCS)
+.PHONY: clean usage help default build fetch $(DEV_DOCS)
```
dblotsky: Nitpick: no newline at end of file.

Makefile
```diff
@@ -50,6 +50,11 @@ else
 printfile = cat $(1)
 endif
 
+ifdef WINDOWS
+RM = cmd /C del /Q /F $(subst /,\,$(1))
+RMDIR  = cmd /C rmdir /S /Q $(subst /,\,$(1))
+endif
```
dblotsky: You can make non-Windows versions of these, and then you won't have to use Windows-specific logic in the `clean` target. So, add an `else` clause and just define:
```makefile
else
RM = rm -f $(1)
RMDIR = rm -rf $(1)
endif
```

Makefile
```diff
@@ -287,6 +292,19 @@ $(CSS_DEST_DIR)/%.css: $(CSS_SRC_DIR)/%.css
...
+	$(call RM, $(DEFAULTS_CONFIG))
+	$(call RM, $(DOCS_PAGE_LIST))
+	$(call RM, $(DOCS_VERSION_DATA))
+	$(call RM, $(TOC_FILES))
+	$(call RM, $(PLUGINS_APP))
+	$(call RM, $(FETCHED_FILES))
+	-$(call RMDIR, $(CSS_DEST_DIR))
+	-$(call RMDIR, $(PROD_DIR))
+	-$(call RMDIR, $(DEV_DIR))
+else
```
dblotsky: If you have `RM` and `RMDIR` defined for all platforms, you don't have to add Windows-specific logic here.

Makefile
```diff
@@ -288,18 +296,19 @@ $(CSS_DEST_DIR)/%.css: $(CSS_SRC_DIR)/%.css
...
+	$(call RM, $(TOC_FILES))
+	$(call RM, $(PLUGINS_APP))
+	$(call RM, $(FETCHED_FILES))
+	-$(call RMDIR, $(CSS_DEST_DIR))
+	-$(call RMDIR, $(PROD_DIR))
+	-$(call RMDIR, $(DEV_DIR))
 
 nuke: clean
 	$(RM) -r node_modules
 	$(RM) Gemfile.lock
```
dblotsky: These also need to be changed to work with the new `RM` and `RMDIR`.

dblotsky:
@gandhirajan these changes look good to me

PR #811, opened 2018-03-28
CB-13955 No mention of 'Make' command running in 'developing-the-website' section

gandhirajan:
Added Make instructions for building and running website


doc/developing-the-website.md
```diff
@@ -11,3 +13,15 @@ That command will build the site and start a local server. To work on only the w
 Alternatively, to dynamically rebuild the site and refresh the browser _when changes happen_ (again, optionally with the `--nodocs` flag), run:
 
     node_modules/.bin/gulp watch
+
+### Using Make
+
+To build the website, run:
+
+    make build
```
janpio: Doesn`t the normal build process have its own document?

janpio: Doesn`t the normal build process have its own document? Or is this required to be able to `serve`?

dblotsky:
I think the Makefile shouldn’t be part of the docs since Gulp is our standard way to do builds. I think it’s best if we document only one way to do builds.

@sujan @stevegill thoughts?

janpio:
IMHO if there are two ways to do things, they should be documented.

doc/developing-the-website.md
gandhirajan: @janpio sorry Jan, I dint get your question. make build should build the website using makefile

dblotsky:
@janpio @stevengill @gandhirajan 

@gandhirajan are you ok with using Gulp? Is there any reason you prefer Make over Gulp?

If not, my recommendation is to delete the Makefile, and to have only one way to do builds. I think two documented ways will be confusing.

gandhirajan:
@dblotsky Hi Dmitry, I have no such preference for Make over Gulp. I just started my cordova contribution with docs and I found this discrepancy and started working on it. If we feel its not required, we can take a collective decision and drop off make. @janpio  @stevengill  your thoughts plz?

doc/developing-the-website.md
dblotsky: @janpio this process is not required at all. The Makefile simply mirrors what the Gulpfile does. I put the Makefile in the project when it was started because I wanted faster builds, but nobody except me liked using Make.

gandhirajan: @dblotsky  @janpio  @stevengill  should we go ahead and knock off the Makefile and its corresponding installation steps then?

purplecabbage:
+1 to removing makefile and leaving documentation explaining the one way to build and serve.

gandhirajan:
@purplecabbage thanks a lot for the response.

dblotsky:
Looks good to me. Farewell, Makefile!

PR #837, opened 2018-06-25
Update docs design to improve readability

Menardi:
### Platforms affected
Docs

### What does this PR do?
Recently, while using the docs, I've found them difficult to navigate and to read. I've made some minor changes to the design in an attempt to make them more readable and easy to use.

I have made the table of contents less wide on larger screens. Additionally I have reduced its padding and changed the styling of the headers to make separation a bit more obvious.

In the main content area, I have increased the line height and header margins to make things a bit easier to read. I have also set a max width on the content to make the text flow better on larger screens.

I've based the changes on docs for popular frameworks such as the [Vue docs](https://vuejs.org/v2/guide/) and [React docs](https://reactjs.org/docs/hello-world.html), as well as the [Onsen UI docs](https://onsen.io/v2/guide/), which I updated recently.

Below are before and after screenshots taken in Chrome set at 1920x1080 resolution. At this resolution, you can see that the text in the content runs on quite long, making it hard to follow.

Before:
Android page before, 1920x1080

After:
Android page after, 1920x1080

And at a lower resolution, 1280x720. In this example, you can see the padding in the table of contents is taking up more space than it should be.

Before:
Android page before, 1280x720

After:
Android page after, 1280x720

Tested on Chrome, Firefox and Safari on macOS, and Edge on Windows.

### Checklist

### DIFF
--- a/www/_layouts/docs.html
+++ b/www/_layouts/docs.html
@@ -25,250 +25,248 @@ NOTE:
 {% endcomment %}
 {% capture PATH_TO_ROOT %}{% for item in my_entry_parts %}{% if forloop.rindex > 1 %}../{% endif %}{% endfor %}{% endcapture %}
 
-<div class="docs container">
+<div class="docs">
 
     <!-- Table of Contents -->
-    <div class="hidden-xs hidden-sm col-md-4 col-lg-3 site-toc-container">
+    <div class="hidden-xs hidden-sm site-toc-container">
         {% include toc_recursive_main.html entries=TOCFILE my_entry=MY_ENTRY path_to_root=PATH_TO_ROOT %}
     </div>
 
     <!-- Page content -->
-    <div class="col-md-8 col-md-offset-4 col-lg-9 col-lg-offset-3 page-content-container container">
-        <div class="page-content row">
-            <div class="col-md-offset-1 col-md-10">
-                <div class="content-header">
+    <div class="page-content-container">
+        <div class="page-content">
+            <div class="content-header">
 
-                    <!-- ToC Dropdown (for XS and SM sizes only) -->
-                    <div class="toc-dropdown dropdown visible-xs-block visible-sm-block">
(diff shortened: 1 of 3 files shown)


janpio:
Looks great!

raphinesse:
I like what you did there! :+1:

But as you'd probably expect from me by now, I still have a few suggestions :wink: 
If we're doing this, I would like to further increase the body type size and further reduce the line length. [Butterick recommends 45-90 characters per line](https://practicaltypography.com/line-length.html). Thus middle ground would be around 68 letters. So the first line could be these 70 characters:

> This guide shows how to set up your SDK environment to deploy Cordova 

I got to the above line length with 18px font size in a 800px container. 800px gets a bit slim for the code samples though, even with the smaller font size that is used there. We should really target 80-120 characters here too, but that's not so easy to change now :smirk:
So from playing around I think 18px body type in a 960px container look reasonable for the main text and the code samples. And it's barely in the recommended range with about 90 characters per line.

Any opinions on this?

Menardi:
I actually originally wanted to do that, but decided to see what the feedback to the other changes would be first. In fact, I looked up articles about line length just like you linked. I think the case is a bit different for technical documentation, compared to an article. I'm not too worried about aiming for the lower end of recommended characters per line.

Increasing the font size is a good way of doing this, rather than making the content container smaller. I've updated the PR (and the screenshots) to set the font size to 18px (16px on mobile). I have also increased the size of the paragraph margin to look better with this increased font size.

Menardi:
@raphinesse @janpio Is there anything to prevent this from being merged? I'm eager to get this merged so the docs are a bit easier to use, since I still refer to them frequently enough :)

janpio:
Nah, looks great as I wrote - I just don't want to be the one merging as it implies a bit of "being responsible" for it and I don't have the time for that right now, sorry.

Menardi:
Ah, that's a shame. Maybe @dpogue or @shazron could help get this merged?

PR #853, opened 2018-07-25
Updated "Sample Usage" in hooks/index.md

AKosmachyov:
### What does this PR do?

Updated "Sample Usage" of hooks.

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/appdev/hooks/index.md
+++ b/www/docs/en/dev/guide/appdev/hooks/index.md
@@ -406,7 +406,7 @@ could be done via hooks.
 ```javascript
 module.exports = function(ctx) {
     // make sure android platform is part of build
-    if (ctx.opts.platforms.indexOf('android') < 0) {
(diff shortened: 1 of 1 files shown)


janpio:
Was this broken before?

raphinesse:
I successfully use `context.opts.platforms` in the `after_prepare` hook of `cordova-plugin-exclude-files`. I wouldn't rule out the possibility that this is different for different types of hooks though. In fact, I think I noticed that once. If that's the case, we should fix the inconsistency.

AKosmachyov:
I've used it with the `after_plugin_install` and got the error "indexOf is not a function"

raphinesse:
Just took a look at the code and can confirm that for `after_plugin_install`, platforms are at `opts.cordova.platforms`. IMHO, that also means that this PR is not the solution to our problem.

Instead, we should consolidate the contexts passed to the different hooks. Anyone know if there's documentation for that?

raphinesse:
Hooks are handled in lib, so this should probably be converted into an issue for lib.

PR #858, opened 2018-07-31
Incorrect default value in config.xml

emanuelecesari:
the default loglevel value is DEBUG not ERROR

### Checklist


janpio:
Thanks for this PR @emanuelecesari 

Do you have a source for this or something so I can understand?

emanuelecesari:
When a new project is created the default value for the loglevel is not **ERROR** as mentioned in the official documentation but is **DEBUG**.
To verify this you can create a new Cordova project and check the _config.xml_ file.
Let me know.

shazron:
Verified. See https://github.com/apache/cordova-android/search?q=loglevel&unscoped_q=loglevel

janpio:
So the default in code is actually `ERROR`, but by default there is a configuration setting that changes it do `DEBUG`?

shazron:
@janpio Looks like it, yes. Default is ERROR https://github.com/apache/cordova-android/blob/ad01d28351c13390aff4549258a0f06882df59f5/framework/src/org/apache/cordova/LOG.java#L38 but it reads the prefs, and sets it to whatever is there, which in this case is DEBUG https://github.com/apache/cordova-android/blob/cc08e9d84a3f7cf76a623006c6a7e48814d74cec/framework/src/org/apache/cordova/CordovaActivity.java#L109

janpio:
Hm, what does "default" in the docs refer to then? Default project? Or default via unconfigured project?
(Follow up question: Should we maybe change the code and config to be the same? Or change the code and remove the default config?)

shazron:
It all depends, do we want to follow the code or the docs? (which one is the truth)
I would fall back on backwards compatibility, in that the truth is the default is DEBUG.

PR #873, opened 2018-08-19
Remove outdated list of supported iOS devices

Menardi:
### Platforms affected
Docs

### What does this PR do?
Removes the list of supported iOS devices from the docs, as it is very outdated. The newest mentioned is the 3-year-old iPhone 6S. Rather than update the list only to see it inevitably get outdated again, I think it's better just to remove it altogether. Cordova works on every iOS device you would want it to, there's no need to list them.

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/ios/index.md
+++ b/www/docs/en/dev/guide/platforms/ios/index.md
@@ -43,12 +43,8 @@ the Apple App Store℠ requires the latest versions of the Apple tools.
 You can test many of the Cordova features using the iOS simulator
 installed with the iOS SDK and Xcode, but you need an actual device to
 fully test all of the app's device features before submitting to the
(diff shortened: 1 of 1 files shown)


Menardi:
I don't think there's use for a table right now, because we haven't dropped an iOS version since cordova-ios@4.0.0 (December 2015). So everyone is likely using a cordova-ios version that only supports iOS 8+, and I don't think anybody will be looking for iOS 7 support! The table is much more useful on Android given the fragmentation of versions, which is less of a problem on iOS.

dpogue:
I believe Apple's official stance is to support only the two most recent versions (so currently iOS 11 and iOS 10, but soon iOS 12 and iOS 11)

Menardi:
Is that to be our official stance? Should we drop support for versions under iOS 11 with the next major (assuming we release it after iOS 12 comes out)?

PR #883, opened 2018-09-12
Full android.util.Config class name for Android

brodybits:
in embedding Android webview doc

to avoid the following potential pitfalls:
- use of "config" with small 'c'
- failure to explicitly import from android.util package

ref:
- <https://stackoverflow.com/questions/45505956/trying-to-embed-cordova-app-into-native-android>
- <https://github.com/apache/cordova-android/issues/494>


brodybits:
Does not look right to me, closing ref: #884

PR #342, opened 2015-09-14
Implemented 404 page with redirect for old docs links

riknoll:
Adds a simple 404 page in desperate need of styling that redirects old documentation links to their new paths (if they exist).

PR #893, opened 2018-09-27
GH-342 add cordova-cli@8.1.1 hotfix blog post

brodybits:


www/_posts/2018-09-27-cordova-cli-8.1.1.md
```diff
@@ -0,0 +1,22 @@
...
+tags: release tools
+---
+
+We are happy to announce that Cordova CLI `8.1.1` hotfix was released in September 2018.
+
+## Release Highlights
+
+The purpose of this hotfix release is to resolve a couple items that were broken in the Cordova CLI `8.1.0` release:
+
+* [GH-339](https://github.com/apache/cordova-cli/issues/339) Fix bin/cordova on Node.js 4
```
janpio: links are to issues, but text seems to be from PRs/commits?

brodybits: Yes. There was one PR with 2 commits to resolve the 2 issues. I can resolve this in one of two ways:
1. Change GH numbers to the PR number and change above text to be more consistent with this solution
2. Change the text to reflect the issue actually solved

I would probably favor the first one. I think it is better for the release note to tell what was actually changed, especially in a hot fix.

brodybits:
@janpio I just made another squash fix for better consistency with the change descriptions.

The one minor downside is that we are using the same GH number for all (both) changes. But I think it is worth it for nicer consistency.

janpio:
I now see GH-X linking to an issue Y - that doesn't make sense.

Just add both links somehow and we will define the actual, useful format in the future for the next releases.

brodybits:
I just reverted the last change, gonna squash-merge it now.

PR #921, opened 2018-12-19
blog post for Cordova node xcode 1.1.0 release

brodybits:


janpio:
Ups, misclicked.

Does this really need a blog post? It is only slightly related to Cordova, not targeted at actual Cordova end users, and also doesn't really change anything.

www/_posts/2018-12-21-xcode-1.1.0.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: post
+author:
+    name: Chris Brody
+    url: https://twitter.com/brodybits
+title:  "Cordova node xcode 8.1.2 Patch Release"
```
erisu: `8.1.2`??? You mean `1.1.0`???

brodybits: Should be fixed now

janpio:
So you just merge the PR instead of answering my question?

brodybits:
> Does this really need a blog post?

I think we should add it since it can slightly change behavior on cordova-ios and cordova-osx scripts.

I think it would be ideal to move cordova-node-xcode out of the Cordova project, just raised <https://github.com/apache/cordova-node-xcode/issues/42>.

Should we consider reverting this one?

PR #971, opened 2019-04-24
Improve Plugin documentation

janpio:
Plugin Development Guide:
- fix git repo url for `cordova plugin add` description
- add npm package name for `cordova plugin add` description
- change explanation of plugin.xml content to list
- fix description of plugin.id
- add several internal links
- move "Validating a Plugin using Plugman" to a better location
- fix formatting

Android Plugin Development Guide:
- Restructure introduction (slightly)
- `src/android/Echo.java` file indentation
- format parameter in text as code

iOS Plugin Development Guide:
- Restructure introduction (slightly)

### DIFF
--- a/www/docs/en/dev/guide/platforms/windows/plugin.md
+++ b/www/docs/en/dev/guide/platforms/windows/plugin.md
@@ -68,6 +68,21 @@ cordova.commandProxy.add("Echo",{
         }
     }
 });
+
+// or alternative syntax
+
+module.exports = {
+    echo: function(successCallback, errorCallback, strInput) {
+        if(!strInput || !strInput.length) {
+            errorCallback("Error, something was wrong with the input string. =>" + strInput);
+        }
+        else {
+            successCallback(strInput + "echo");
+        }
+    }
+};
+
+require('cordova/exec/proxy').add('Echo', module.exports);
 ```
 
 The `www/echoplugin.js` file will forward the `echo` function call to this proxy through the `cordova.exec` command and execute this implementation in `src/windows/echopluginProxy.js`.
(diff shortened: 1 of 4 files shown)


timbru31:
Thanks for updating the documentation! Getting started with plugin development was really a pain so far ;)

janpio:
I feel you, just going through the process myself again and taking the opportunity to make this a bit less terrible.

But sounds like you did the same @timbru31 - do you have any stuff to add here?

TODO from Slack:
- `--link` (added via #975) only applies to native parts of the plugin (not package.json, .js etc) and so is only sometimes useful ("the only thing is that changes in javascript, types, plugin.xml or package.json are not propagated, just the native files")
- `--link` workflow: create the project, add the plugin via `--link`, then run with xcode/androidstudio, modify native parts of the plugin in the plugin source folder, re-run.

timbru31:
> do you have any stuff to add here?

Sure. Will add some input later :)

www/docs/en/dev/guide/platforms/windows/plugin.md
```diff
@@ -66,11 +68,26 @@ cordova.commandProxy.add("Echo",{
         }
     }
 });
+
+// or alternative syntax
```
janpio: No idea if this actually works - will have to try it and compare that to the currently documented implementation.

janpio: Those are probably both actually calling https://github.com/apache/cordova-js/blob/a770d95f129f0e16c696bcecd00dc8d3e8e53b5c/src/common/exec/proxy.js

janpio:
TODO

"Plugin Exec Proxy" form the WIndows plugin docs also applies to browser and can be used for all - maybe move description to general plugin docs file

erisu:
@timbru31 It looks like you were going to leave some input? If not and if you feel this is OK as it stands, I like to merge it in.

timbru31:
I’ll add this to my to-do list for tomorrow. If I fail to do so, feel free to merge on Monday

PR #976, opened 2019-04-27
cordova-windows 6+: Windows 10 default

janpio:
This updates the Windows docs for what is reality since cordova-windows 6.x


www/docs/en/dev/guide/platforms/windows/index.md
```diff
@@ -24,15 +24,15 @@ toc_title: Windows
...
+and deploy Cordova apps Windows 10 (Universal Windows Platform [= UWP], formerly known as Universal App Platform [= UAP]), 
+Windows 8.1 and Windows Phone 8.1.  It shows how to use either shell tools
 to generate and build apps, or the cross-platform Cordova CLI. (See the [Overview](../../overview/index.html#development-paths) for a comparison of these
 development options.) This section also shows how to modify Cordova apps
 within Visual Studio. Regardless of [which](../../overview/index.html#development-paths) approach you take, you need to
 install the Visual Studio SDK, as described below.
 
-Developers wishing to target Windows Phone 8 should use the wp8 platform,
-see [Windows Phone 8 Platform Guide](../wp8/index.html) for details (Warning, the wp8 platform is deprecated).
+Note: Developers wishing to target Windows Phone 8 (instead of 8.1!) should use the `wp8` platform,
```
erisu: IMO it is not necessary to talk about or suggest to users to use a deprecated platform.

www/docs/en/dev/guide/platforms/windows/index.md
```diff
@@ -90,17 +84,31 @@ The tools and SDKs for the target Windows platforms (UWP, 8.1, etc.) must also b
...
+You may decide that you want to build a particular version of your application targeting a particular OS (for example, you might have set that you want to target Windows 10, but you want to build for Windows Phone 8.1).  To do this, you can use the `--appx` parameter:
+
 ```
+cordova build windows -- --appx=8.1-phone
+```
+
+The build system will ignore the preference set in config.xml for the target Windows version and strictly build a package for Windows Phone 8.1.
 
-Once you add this setting `build` command will start producing Windows 10 packages.
+Valid values for the `--appx` flag are `8.1-win`, `8.1-phone`, and `uwp` or `uap` (for Windows 10 Universal Apps / Universal Windows Apps).  These options also apply to the `cordova run` command.
```
erisu: Maybe convert this to a table?? Other then that this is OK still.

janpio: Not worth the effort as all the 8.1 stuff will be gone soon - and as such probably the complete `appx` param. Otherwise, yeah.

www/docs/en/dev/guide/platforms/windows/index.md
janpio: As I understand this this is historically as both grew from the same platform. But this change just clarifies this existing sentence - didn't just want to remove it.

janpio: Oh, that platform does not exist any more at all in the docs - so this indeed does make no sense to mention here this way.

janpio: Removed.

www/docs/en/dev/guide/platforms/windows/index.md
```diff
@@ -23,55 +23,30 @@ toc_title: Windows
...
-- A Windows 8.1, 32 or 64-bit machine (_Home_, _Pro_, or _Enterprise_ editions)
-  with minimum 4 GB of RAM along with [Visual Studio 2015](http://www.visualstudio.com/downloads)
-  or Visual Studio 2013.  An evaluation version of Windows 8.1 Enterprise is
-  available from the
-  [Microsoft Developer Network](https://technet.microsoft.com/evalcenter/hh699156.aspx).
+To develop apps for Windows 10 you need:
 
-- For the Windows Phone emulators, Windows 8.1 (x64) Professional edition or higher,
-and a processor that supports <a href='https://msdn.microsoft.com/en-us/library/windows/apps/ff626524(v=vs.105).aspx#hyperv'>Client Hyper-V and Second Level Address Translation (SLAT)</a>.
+- Windows 10 (or Windows 8.1), 32- or 64-bit, along with [Visual Studio 2015](http://www.visualstudio.com/downloads) or higher.
```
brodybits: I think Cordova is not directly supported with Visual Studio 2019 since it dropped .jsproj support, as discussed in apache/cordova-windows#327.

janpio: Good point, we can use this PR to include this information as well.

janpio: Fixed and specified.

www/docs/en/dev/config_ref/index.md
```diff
@@ -391,9 +391,8 @@ Examples:
 <preference name="AppendUserAgent" value="My Browser" />
 
 <!-- Windows only preferences -->
-<preference name="windows-phone-target-version" value="8.1" />
 <preference name="windows-target-version" value="8.1" />
-<preference name="Windows.Universal" value="10.0.10240.0" />
```
janpio: From my understanding of https://github.com/apache/cordova-windows/blob/8cc627b8f5fb4a0959c4ef7a5994a4f358215fd6/template/cordova/lib/ConfigParser.js#L103-L170 this could never actually work and was probably a misunderstanding how the preference actually worked (what was written in this line is the fallback that is used if nothing is supplied here). Lines 398 and 399 show two working examples.

PR #977, opened 2019-04-28
doc: extract project structure into its own file

janpio:
This file is extraxcted from cordova-cli via https://github.com/apache/cordova-cli/pull/428

Slight modifications already done, will have to checked if still up to date and probably restructured quite a bit.

Comments and suggestions welcome!

### DIFF
--- a/www/docs/en/dev/guide/overview/project-structure.md
+++ b/www/docs/en/dev/guide/overview/project-structure.md
@@ -19,40 +19,30 @@ license: >
 
 title: Project Structure of a Cordova Project
 toc_title: Project Structure
-description: 
+description:
 ---
 
 # Project Structure
 
-## CLI project directory structure
+## CLI's Default Directory Structure
 
-Cordova CLI works with the following directory structure:
+A project created with Cordova CLI comes with the follow directory structure by default:
 
-```
+```text
 myapp/
 |-- package.json
 |-- config.xml
-|-- hooks/
-|-- merges/
-| | |-- android/
-| | |-- windows/
-| | |-- ios/
 |-- node_modules/
 |-- www/
 |-- platforms/
-| |-- android/
-| |-- windows/
-| |-- ios/
 |-- plugins/
-  |--cordova-plugin-camera/
-  
 ```
-TODO resort properly
 
+TODO resort properly
 
 ### `package.json`
 
-TODO 
+TODO
 
 ### `config.xml`
 
@@ -65,7 +55,7 @@ Contains the project's web artifacts, such as .html, .css and .js files. As a co
 ### `node_modules/`
 
 TODO
-Dependencies inistalled via NPM. 
+Dependencies inistalled via NPM.
 Base for `platforms` and `plugins`
 
 ### `platforms/`
@@ -78,17 +68,17 @@ Contains all the source code and build scripts for the platforms that you add to
 
 Any added plugins will be extracted or copied into this directory.
 
-### `hooks/`
+### Version control
 
-This directory may contains scripts used to customize cordova-cli commands. Any scripts you add to these directories will be executed before and after the commands corresponding to the directory name. Useful for integrating your own build systems or integrating with version control systems.
+It is recommended not to check in `platforms/` and `plugins/` directories into version control as they are considered a build artifact. Your platforms and plugins will be saved in `package.json` automatically. These platforms/plugins will be downloaded when `cordova prepare` is invoked.
 
-Refer to [Hooks Guide] for more information.
+## Optional Directories
 
 ### `merges/`
 
 Platform-specific web assets (HTML, CSS and JavaScript files) are contained within appropriate subfolders in this directory. These are deployed during a `prepare` to the appropriate native directory.  Files placed under `merges/` will override matching files in the `www/` folder for the relevant platform. A quick example, assuming a project structure of:
 
-```
+```text
 merges/
 |-- ios/
 | -- app.js
@@ -99,7 +89,3 @@ www/
 ```
 
 After building the Android and iOS projects, the Android application will contain both `app.js` and `android.js`. However, the iOS application will only contain an `app.js`, and it will be the one from `merges/ios/app.js`, overriding the "common" `app.js` located inside `www/`.
-
-### Version control
-
-It is recommended not to check in `platforms/` and `plugins/` directories into version control as they are considered a build artifact. Your platforms and plugins will be saved in `package.json` automatically. These platforms/plugins will be downloaded when `cordova prepare` is invoked.
--- a/www/docs/en/dev/guide/overview/project-structure.md
+++ b/www/docs/en/dev/guide/overview/project-structure.md
@@ -17,9 +17,9 @@ license: >
     specific language governing permissions and limitations
     under the License.
 
-title: Architectural overview of Cordova platform
-toc_title: Overview
-description: Start here if you are new to Cordova. Includes installation and next steps.
+title: Project Structure of a Cordova Project
+toc_title: Project Structure
+description: 
 ---
 
 # Project Structure
--- a/www/docs/en/dev/guide/overview/project-structure.md
+++ b/www/docs/en/dev/guide/overview/project-structure.md
@@ -37,6 +37,7 @@ myapp/
 | | |-- android/
 | | |-- windows/
 | | |-- ios/
+|-- node_modules/
 |-- www/
 |-- platforms/
 | |-- android/
@@ -44,7 +45,10 @@ myapp/
 | |-- ios/
 |-- plugins/
   |--cordova-plugin-camera/
+  
 ```
+TODO resort properly
+
 
 ### `package.json`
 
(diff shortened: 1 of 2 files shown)


www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,105 @@
...
+| -- android.js
+www/
+-- app.js
+```
+
+After building the Android and iOS projects, the Android application will contain both `app.js` and `android.js`. However, the iOS application will only contain an `app.js`, and it will be the one from `merges/ios/app.js`, overriding the "common" `app.js` located inside `www/`.
+
+### Version control
+
+It is recommended not to check in `platforms/` and `plugins/` directories into version control as they are considered a build artifact. Your platforms and plugins will be saved in config.xml & package.json automatically. These platforms/plugins will be downloaded when `cordova prepare` is invoked.
```
dpogue: This should say "Your platforms and plugins will be saved in package.json automatically."

We no longer save them into config.xml as of Cordova 9.

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,105 @@
...
+## CLI project directory structure
+
+Cordova CLI works with the following directory structure:
+
+```
+myapp/
+|-- package.json
+|-- config.xml
+|-- hooks/
+|-- merges/
```
dpogue: The merges folder is not created automatically anymore, and we kinda discourage its use (although it's still supported as far as I know).

Also, the example folders here are indented one level too far.

janpio: Is what is written below about `merges` still accurate though?
Will have to splti existing document into "created by default" and "also supported" somehow I guess.

www/docs/en/dev/guide/overview/project-structure.md
janpio: Yeah, exactly why I created this PR. This information being buried in the CLI referenc means it is ignored and I think most users never have seen it - but it could be so useful to understand what Cordova is actually doing and how it works.

I would love if you could add more comment on inaccuracies or missing information of this type here. Will try to work it all in.

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,105 @@
...
+
+## CLI project directory structure
+
+Cordova CLI works with the following directory structure:
+
+```
+myapp/
+|-- package.json
+|-- config.xml
+|-- hooks/
```
raphinesse: As with merges, usage of the `hooks` folder is deprecated.

janpio: Hooks are still supported via config files though, and then can be in random locations, right?

So this here should be removed, but maybe a new paragraph that links to the hooks documentation (that hopyfully exists).

dpogue: Yes. Hooks can live anywhere, and we want them explicitly listed in config.xml now instead of auto-detected through the hooks folder.

janpio: For later reference: https://cordova.apache.org/docs/en/9.x/guide/appdev/hooks/index.html is the docs about hooks

erisu: I applied changes so this can be re-reviewed.

www/docs/en/dev/guide/overview/project-structure.md
erisu: I applied changes so this can be re-reviewed.

I tried to reorg the structure so there is now the "default" and "optional" section.

www/docs/en/dev/guide/overview/project-structure.md
erisu: I applied changes so this can be re-reviewed.

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+
+TODO
+
+### `config.xml`
+
+Configures your application and allows you to customize the behavior of your project. See also [config.xml reference documentation][config.xml ref]
+
+### `www/`
+
+Contains the project's web artifacts, such as .html, .css and .js files. As a cordova application developer, most of your code and assets will go here. They will be copied on a `cordova prepare` to each platform's www directory. The www source directory is reproduced within each platform's subdirectory, appearing for example in `platforms/ios/www` or `platforms/android/assets/www`. Because the CLI constantly copies over files from the source www folder, you should only edit these files and not the ones located under the platforms subdirectories. If you use version control software, you should add this source www folder, along with the merges folder, to your version control system.
```
dpogue, line 51:
```suggestion
Contains the project's web artifacts, such as HTML, CSS, JavaScript, and other resource asset files.

As a Cordova application developer, most of your code and assets will go here. They will be copied on a `cordova prepare` to each platform's `www` directory. The `www` source directory is reproduced within each platform's subdirectory, appearing for example in `platforms/ios/www` or `platforms/android/assets/www`. Because the CLI constantly copies over files from the source `www` folder, you should only edit these files and not the ones located under the platforms subdirectories.

If you use a JavaScript build tool, you should set it to output your production distribution files to the `www` folder.
If you are developing code in the `www` folder directly, you should add this  folder to your version control system.
```

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+
+### `node_modules/`
+
+This directory contains all of the checked-out packages from the npmjs registry.
```
dpogue, line 55:
```suggestion
This directory contains all of the JavaScript dependency packages from the npm JavaScript registry for Cordova and its tools, along with any dependencies of your project (as specified in `package.json`).
```

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+### `platforms/`
+
+Contains all the source code and build scripts for the platforms that you add to your project.
+
+> **WARNING:** When using the CLI to build your application, you should not edit any files in the /platforms/ directory unless you know what you are doing, or if documentation specifies otherwise. The files in this directory are routinely overwritten when preparing applications for building, or when plugins are re-installed.
+
+### `plugins/`
+
+Any added plugins will be extracted or copied into this directory.
+
```
dpogue, line 76: Do we want the same warning here about not checking in the `/plugins` folder to version control?

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+├── node_modules/
+├── package.json
+├── platforms/
+└── plugins/
+└── www
+```
+
+### `package.json`
+
+TODO
```
dpogue, line 43:
```suggestion
A manifest declaring what JavaScript package dependencies are used by your project, including Cordova, the added platforms, and any added plugins. This is also where plugin variable values are stored.

Your project might already have a `package.json` file with its own dependencies, or one generated from a framework. Cordova will simply add what it needs to the `package.json` file without interfering with other tools or dependencies.
```

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+
+### `config.xml`
+
+Configures your application and allows you to customize the behavior of your project. See also [config.xml reference documentation][config.xml ref]
```
dpogue, line 47:
```suggestion
Contains the preferences and configuration options for your Cordova application and allows you to customize the behavior of your project.

See also [`config.xml` reference documentation][config.xml ref].
```

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,100 @@
...
+
+A project created with Cordova CLI comes with the follow directory structure by default:
+
+```text
+myapp/
+├── config.xml
+├── node_modules/
+├── package.json
+├── platforms/
+└── plugins/
```
dpogue, line 37:
```suggestion
├── plugins/
```

www/docs/en/dev/guide/overview/project-structure.md
erisu: Yeah, I think a warning should be added to not commit the plugins directory to version control.

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,109 @@
...
+
+### `platforms/`
+
+Contains all the source code and build scripts for the platforms that you add to your project.
+
+> **WARNING:** When using the CLI to build your application, you should not edit any files in the /platforms/ directory unless you know what you are doing, or if documentation specifies otherwise. The files in this directory are routinely overwritten when preparing applications for building, or when plugins are re-installed.
+
+### `plugins/`
+
+Any added plugins will be extracted or copied into this directory.
```
erisu, line 84:
```suggestion
The `plugins` directory is used as a staging area where installed plugins are copied before being applied to the `platforms` directory.

This directory should not be committed to version control.

If you've created custom, in-house or modified plugins and want to store them within the project, it's recommended **not** to store them in the `plugins` directory.
```

erisu (approved):

www/docs/en/dev/guide/overview/project-structure.md
```diff
@@ -0,0 +1,113 @@
...
+
+A manifest declaring what JavaScript package dependencies are used by your project, including Cordova, the added platforms, and any added plugins. This is also where plugin variable values are stored.
+
+Your project might already have a `package.json` file with its own dependencies, or one generated from a framework. Cordova will simply add what it needs to the `package.json` file without interfering with other tools or dependencies.
+
+### `config.xml`
+
+Contains the preferences and configuration options for your Cordova application and allows you to customize the behavior of your project.
+
+See also [`config.xml` reference documentation][config.xml ref].
```
dpogue, line 51: This link doesn't seem to have its target "config.xml ref" defined anywhere

dpogue (approved):

PR #985, opened 2019-05-08
Faster PR builds & use Travis CI's deploy phase

### LINKED ISSUE #981: buildAndDeploy.sh does not fail Travis CI build when something goes wrong
https://github.com/apache/cordova-docs/issues/959 shows that a failure during the deploy process via https://github.com/apache/cordova-docs/blob/master/buildAndDeploy.sh does not fail the build if something goes wrong.

Example:
"Successful" build at:
https://travis-ci.org/apache/cordova-docs/builds/529733608?utm_source=github_status&utm_medium=notification
If actually look at the raw log:
https://api.travis-ci.org/v3/job/529733609/log.txt
You see this:
```
svn: E215004: Authentication failed and interactive prompting is disabled; see the --force-interactive option
svn: E215004: Commit failed (details follow):
svn: E215004: No more credentials or we tried too many times.
Authentication failed
```

So clearly the `svn commit` actually failed, but the failure was not reported to Travis to alert us about this problem.

raphinesse:
### Motivation and Context

CI was slow and buggy.
Fixes #981 

### Description

- Always run `npm test` during CI runs
- Do a quick development build during `npm test`
- Only run full production build before deployment
- Replace `buildAndDeploy.sh` by `deploy.sh` which deploys an existing
  production build and is also usable outside of a Travis CI context
- Unchanged: Only deploy on commits to master (not when checking PRs)

### Testing

I tested correct operation of the new Travis CI config and the deploy script on my fork. The only thing untested is the last line in `deploy.sh`, the actual commit.


.travis.yml
```diff
@@ -1,12 +1,17 @@
 language: node_js
+node_js: 8
 sudo: false
-node_js:
-- '8'
```
dblotsky: Does changing this from a list to an int continue to work in Travis CI?

tools/bin/deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+        echo 'ERROR: No SVN credentials given in $svn_username and $svn_pass'
+        exit 1
+    }
+    svn_q --non-interactive --no-auth-cache \
+          --username="$svn_username" --password="$svn_pass" "$@"
+}
+
+echo "Deploying website and docs to $svn_url"
+cd ..
+svn_q checkout $svn_url cordova-website
```
dblotsky: Please put the URL in quotes.

package.json
```diff
@@ -7,7 +7,7 @@
     "doc": "doc"
   },
   "scripts": {
-    "test": "npm run eslint && npm run mocha",
+    "test": "npm run eslint && npm run mocha && gulp build",
```
dblotsky: Does any step of the deployment process run this?

.travis.yml
```diff
@@ -1,12 +1,17 @@
...
-- bundle install
+
 install:
-- npm install
-script:
-  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
-  - chmod 755 buildAndDeploy.sh
-  - travis_wait 30 sleep infinity & ./buildAndDeploy.sh
+  - npm install
+  - bundle install
```
dblotsky: Might this be a good place to add a `- npm test`?

dblotsky:
Thanks for the PR! A more philosophical question: what do you think of moving the `npm run *` commands to `gulpfile.js`?

.travis.yml
erisu: haha nice catch, I didn't even noticed the test command was not even used.

raphinesse: `npm test` is the default `script`, i.e. main CI phase for projects with `language: node_js`

package.json
raphinesse: This will run before each _continuous_ deployment by Travis CI (also see below). The CI run will not continue with deployment if the `npm test` run fails. Running only the deploy script yourself will not run this.

tools/bin/deploy.sh
raphinesse: I thought it will probably never contain spaces, but I know myself that that's a weak argument. Will do.

raphinesse: Yes. Also see the CI run for this PR.

.travis.yml
raphinesse: Yes. Also see the CI run for this PR.

raphinesse:
> Thanks for the PR! A more philosophical question: what do you think of moving the `npm run *` commands to `gulpfile.js`?

IMHO all important operations on a `npm` package should be accessible by an entry in `scripts`. The nice thing is that it sets up your `PATH` to include your dependencies' binaries etc. That way you do not have to install `gulp-cli` in this case. Of course you could also use `npx` for that, but the `scripts` give a nice list of what you are supposed to be doing. Moreover, Travis CI runs `npm test` for you by default :grin: 

Gulp OTOH is fine for complex build steps, just like it is used here. I wouldn't use it as an entry point to all and everything.

.travis.yml
janpio: I would prefer to make it explicit in the CI configuration, people changing this in the future (looking in the mirror: me!) might now know about it and wonder what is going on. Doesn't hurt having it, does it?

package.json
janpio: What does `gulp build` do?

janpio: Ah, this is:
> Do a quick development build during npm test

.travis.yml
raphinesse: It hurts my eyes! :scream:

JK, it's definitely more obvious with the `script` phase explicitly defined.

Issue #838, opened 2018-06-26
Docs should not encourage global npm installation


dpogue:
Currently the docs say to run `sudo npm install -g cordova` which is not great from a best practices standpoint. We should never encourage running npm with sudo because of permissions problems, and npm recommends not installing anything globally if it can be helped.

We should update our documentation to use `npx cordova` instead ([npx](https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner) requires npm >=5.2.0), and add cordova as a devDependency to our hello-world project template.

brodybits:
> Currently the docs say to run `sudo npm install -g` cordova which is not great

+1 (+100 if you change "not great" to "bad")

> use `npx cordova` instead

- does not work for `create`
- I could image a developer wanting to use own installation of Cordova CLI, regardless of what is used by the project

I would favor an approach inspired by React Native:
- Simplistic create Cordova app tool
- The generated Cordova app could have npm scripts for some the most important tasks from Cordova CLI

I guess the user could have a choice whether to use "npx cordova" or own installation of Cordova CLI for other tasks.

dpogue:
It works perfectly for create: `npx cordova create MyProject`
You can even use a specific version: `npx cordova@5.4.0 create MyProject`

raphinesse:
I'd suggest presenting both alternatives:

- global installation (without using sudo)
- Add to devDeps and use with npx (advertise as preferred!)

And then just keep using plain `cordova` throughout the docs.

janpio:
There probably was a reason why `sudo` is mentioned currently?

> And then just keep using plain cordova throughout the docs.

Will this work on all platforms without the global `cordova` install?

dpogue:
>> And then just keep using plain cordova throughout the docs.
>
> Will this work on all platforms without the global cordova install?

No, if you don't install cordova globally, you'd always need to prefix it with `npx` when you want to run it.

`npx` will look inside node_modules for it, and use it from there if it exists, or else it will install it into a temporary folder and run it from there for the duration of the command

raphinesse:
> There probably was a reason why `sudo` is mentioned currently?

@janpio I suppose because for most default setups you need root access to install npm packages globally. This is widely considered a bad idea though. Instead you should setup npm to install global packages to your home folder, for example. Other than that, `sudo` is still mainly an Ubuntu thing, or am I mistaken?

Anyway, I was suggesting the following:

Inform the user *once* about his options on how to install `cordova`:
[...]

PR #987, opened 2019-05-09
Prefer local cordova CLI with npx

oliversalzburg:
### Platforms affected

None

### Motivation and Context

See #838 for discussion about this change.

### Description

This changes the documentation to recommend using a project-local installation of `cordova`, instead of relying on a global installation.

The thinking that went into this change was:

- When a command or script is simply mentioned, to refer to the task it handles, no `npx` should **not** be prefixed.
- If the given command is an example of invocation, `npx` should be prefixed.

I tried to be consistent with this, hopefully, I didn't fail.

I also tried to bring more consistency to syntax highlighting and prefixing CLI usages with `$`. I hope that wasn't completely out of scope for this change.

### Checklist

### DIFF
--- a/www/contribute/issues.md
+++ b/www/contribute/issues.md
@@ -32,15 +32,27 @@ If you aren't sure about the priority, leave the default (major) selected. Pleas
 
 You can quickly find out versions of platforms/plugins you're using by running:
 
-    cordova platform ls
+```bash
+$ npx cordova platform ls
+```
 
 or
 
-    cordova plugin ls
+```bash
+$ npx cordova plugin ls
+```
 
 in your project respectively. You can find out the version of the Cordova CLI you're using by running:
 
-    cordova --version
+```bash
+$ npx cordova --version
+```
+
+If you have installed `cordova` globally (which is an alternative way to use Cordova), you just invoke the same commands without `npx`:
+
+```bash
+$ cordova --version
+```
 
 # Generating CLI logs
 
--- a/www/docs/en/dev/guide/cli/index.md
+++ b/www/docs/en/dev/guide/cli/index.md
@@ -35,59 +35,22 @@ The Cordova command-line tool is distributed as an npm package.
 To install the `cordova` command-line tool, follow these steps:
 
 1. Download and install [Node.js](https://nodejs.org/en/download/). On
-   installation you should be able to invoke `node` and `npm` on your
+   installation you should be able to invoke `node`, `npm` and `npx` on your
    command line.
 
 1. (Optional) Download and install a [git client](http://git-scm.com/downloads), if you don't
    already have one. Following installation, you should be able to invoke `git`
    on your command line. The CLI uses it to download assets when they are referenced using a url to a git repo.
 
-1. Install the `cordova` module using `npm` utility of Node.js. The `cordova`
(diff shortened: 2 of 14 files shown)


oliversalzburg:
The `www/docs/en/dev/guide/appdev/hooks/index.md` have a couple of use cases I was not yet sure on how they should look now. I also need to check how `CORDOVA_CMDLINE` behaves with this approach.

www/contribute/issues.md
```diff
@@ -32,14 +32,24 @@ If you aren't sure about the priority, leave the default (major) selected. Pleas
...
+$ npx cordova plugin ls
+```
 
 in your project respectively. You can find out the version of the Cordova CLI you're using by running:
 
+```bash
+$ npx cordova --version
+```
+
+If you have installed `cordova` globally (which is not recommended), you just invoke the same commands without `npx`:
```
janpio: "not recommended" is strong wording. We recommended exactly this for ~10 years. "alternative way to use Cordova" seems more suitable.

dblotsky:
I don’t think recommending `npx` is responsible advice to our users; more details in #838.

brodybits:
> I also need to check how `CORDOVA_CMDLINE` behaves with this approach.

Can you elaborate on this?

I wonder if this should be discussed in a separate issue or mailing list thread?

(I would favor a separate issue that is linked to this PR for easier tracking.)

oliversalzburg:
The documentation says that `CORDOVA_CMDLINE` is set to the command line with which cordova was invoked. If this includes `npx` when it is invoked with `npx`, what does that imply in terms of compatibility? I'm guessing, nothing, because it was a working approach to invoking cordova previously, but I wanted to look into it anyway.

www/docs/en/dev/guide/cli/index.md
```diff
@@ -35,78 +35,54 @@ The Cordova command-line tool is distributed as an npm package.
...
+```bash
+$ cd hello
+```
 
-Go to the directory where you maintain your source code, and create a cordova project:
+Until necessary adjustments have been made to Cordova to do this for your automatically, you're going to want to install the current version of Cordova into your new project:
 
 ```bash
-$ cordova create hello com.example.hello HelloWorld
+$ npm --save-dev cordova
```
purplecabbage: `$ npm install --save-dev cordova`

www/docs/en/dev/guide/cli/index.md
```diff
@@ -35,78 +35,54 @@ The Cordova command-line tool is distributed as an npm package.
...
+All subsequent commands need to be run within the project's directory,
+or any subdirectories:
 
-## Create the App
+```bash
+$ cd hello
+```
 
-Go to the directory where you maintain your source code, and create a cordova project:
+Until necessary adjustments have been made to Cordova to do this for your automatically, you're going to want to install the current version of Cordova into your new project:
```
janpio:
```suggestion
Until necessary adjustments have been made to Cordova to do this for you automatically, you're going to want to install the current version of Cordova into your new project:
```

purplecabbage:
Closing this as stale.
I think the use case is still valid, but this should be presented as an alternative way of setting up a project, and not go through and replace all existing commands with npx versions.

PR #990, opened 2019-05-10
Update to latest gulp-sass

raphinesse:
This has the following benefits:

- Uses a node-sass version that officially supports Node.js 8
- Speeds up installation because we do not need to compile node-sass
- Fixes some npm security audit warnings

This slightly changes the _formatting_ of the website's generated CSS.
Other than that, this does not change the build output.


dblotsky:
Please take a look at the follow-up #993.

Issue #986, opened 2019-05-09
Explain platform version pinning


janpio:
Our platforms are pinned, so that `cordova platform add ...` installs a specific version. This should be explained in the docs somewhere, so that cases like this can link to that explanation: https://github.com/apache/cordova-docs/pull/979#discussion_r282398059

PR #991, opened 2019-05-10
Added Platform Pinning Docs

erisu:
### Platforms affected
none

### Motivation and Context
Talks about platform pinning.

closes: #986

### Description
Added docs

### Testing
- Ran server

### Checklist

### DIFF
--- a/www/docs/en/9.x/platform_pinning/index.md
+++ b/www/docs/en/9.x/platform_pinning/index.md
@@ -1,0 +1,50 @@
+---
+license: >
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+title: Platform Pinning
+toc_title: Platform pinning
+description: Cordova CLI pinned platform versions.
+---
+
+## Platform Pinning
+
+Cordova CLI pins the supported platform version in conjunction with the version of CLI that is being used.
+
+To see the pinned platforms for your CLI version, execute the command `cordova platform list`  in a brand new project directory.
+
+**Example (Cordova CLI 9.x):**
+
+```bash
+$ cordova platform list
+Installed platforms:
+
+Available platforms:
+  android ^8.0.0
+  browser ^6.0.0
+  electron ^1.0.0
+  ios ^5.0.0
+  osx ^5.0.0
+  windows ^7.0.0
+```
+
+_Note: When a platform is installed, the "**Installed platforms:**" section will display the actual installed platform version. The installed platform will no longer display in the "**Available platforms:**" section until it is removed from the project._
+
+Starting from **Cordova CLI 9** and up, all platforms are pinned with an caret (`^`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a minor release.
+
+In **Cordova CLI 8** and below, all platforms are pinned with a tilde (`~`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a patch release.
--- a/www/docs/en/dev/platform_pinning/index.md
+++ b/www/docs/en/dev/platform_pinning/index.md
@@ -1,0 +1,50 @@
+---
+license: >
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+title: Platform Pinning
+toc_title: Platform pinning
+description: Cordova CLI pinned platform versions.
+---
+
+## Platform Pinning
+
+Cordova CLI pins the supported platform version in conjunction with the version of CLI that is being used.
+
+To see the pinned platforms for your CLI version, execute the command `cordova platform list`  in a brand new project directory.
+
+**Example (Cordova CLI 9.x):**
+
+```bash
+$ cordova platform list
+Installed platforms:
+
+Available platforms:
+  android ^8.0.0
+  browser ^6.0.0
+  electron ^1.0.0
+  ios ^5.0.0
+  osx ^5.0.0
+  windows ^7.0.0
+```
+
+_Note: When a platform is installed, the "**Installed platforms:**" section will display the actual installed platform version. The installed platform will no longer display in the "**Available platforms:**" section until it is removed from the project._
+
+Starting from **Cordova CLI 9** and up, all platforms are pinned with an caret (`^`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a minor release.
+
+In **Cordova CLI 8** and below, all platforms are pinned with a tilde (`~`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a patch release.
--- a/www/docs/en/dev/platform_pinning/index.md
+++ b/www/docs/en/dev/platform_pinning/index.md
@@ -24,11 +24,11 @@ description: Cordova CLI pinned platform versions.
 
 ## Platform Pinning
 
-Cordova CLI pins the supported platform version in conjunction with the version of CLI that is being used.
+Cordova CLI manages a list of pinned Apache Cordova maintained platforms. When there is a major release of Cordova CLI, the platform's pinned version is updated to the latest major released version. Typically, the updating of this pin is made only on a major release of CLI. This is because the pinned versions are pinned with a caret `^` allowing CLI to continue to fetch new minor and patch releases of any given pinned platform.
 
 To see the pinned platforms for your CLI version, execute the command `cordova platform list`  in a brand new project directory.
 
-**Example (Cordova CLI 9.x):**
+**Cordova CLI 9.x:**
 
 ```bash
 $ cordova platform list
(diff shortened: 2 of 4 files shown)


www/docs/en/9.x/platform_pinning/index.md
```diff
@@ -0,0 +1,52 @@
...
+  browser ^6.0.0
+  electron ^1.0.0
+  ios ^5.0.0
+  osx ^5.0.0
+  windows ^7.0.0
+```
+
+_Note: When a platform is installed, the "**Installed platforms:**" section will display the actual installed platform version. The installed platform will no longer display in the "**Available platforms:**" section until it is removed from the project._
+
+Starting from **Cordova CLI 9** and up, all platforms are pinned with an caret (`^`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a minor release.
```
janpio: add example at end here

www/docs/en/9.x/platform_pinning/index.md
```diff
@@ -0,0 +1,52 @@
...
+    under the License.
+
+title: Platform Pinning
+toc_title: Platform pinning
+description: Cordova CLI pinned platform versions.
+---
+
+## Platform Pinning
+
+Cordova CLI pins the supported platform version in conjunction with the version of CLI that is being used.
```
janpio: complicated. better way to say "each CLI version has a list of platform versions pinned in a file"?

www/docs/en/9.x/platform_pinning/index.md
```diff
@@ -0,0 +1,52 @@
...
+  ios ^5.0.0
+  osx ^5.0.0
+  windows ^7.0.0
+```
+
+_Note: When a platform is installed, the "**Installed platforms:**" section will display the actual installed platform version. The installed platform will no longer display in the "**Available platforms:**" section until it is removed from the project._
+
+Starting from **Cordova CLI 9** and up, all platforms are pinned with an caret (`^`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a minor release.
+
+In **Cordova CLI 8** and below, all platforms are pinned with a tilde (`~`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a patch release.
```
janpio: add example

www/docs/en/9.x/platform_pinning/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+_Note: It is possible to install a newer version of a platform, even when it is not listed as a supported version. However, it is highly not recommended._
```
janpio: hu?

www/docs/en/9.x/platform_pinning/index.md
erisu: If written this way, I suspect users would want to see a link to this file?

The pinning is actually [here](https://github.com/apache/cordova-lib/blob/master/src/platforms/platformsConfig.json) in `cordova-lib`.

Any changes to make it less complicated is great, I only wanted to point out that if we start adding links to files in code, it means we would have to maintain this more frequently. 

On every release, we would need to update the link to the correct release version (not master) in LIB and anyone who does the future release of lib that includes pinning changes, they must know about this page and that it should as well be updated.

www/docs/en/9.x/platform_pinning/index.md
erisu: Do you mean move the `**Example (Cordova CLI 9.x):**` section below `Starting from **Cordova CLI 9**  and up` sentence?

www/docs/en/9.x/platform_pinning/index.md
erisu: Do you mean display below `In **Cordova CLI 8** and below `  something like this:

```
**Example (Cordova CLI 8.x):**

```bash
$ cordova platform list
Installed platforms:

Available platforms:
  android ~7.1.1
  browser ~5.0.1
  ios ~4.5.4
  osx ~4.0.1
  windows ~6.0.0
```

www/docs/en/9.x/platform_pinning/index.md
erisu: Example: **Cordova CLI 9.x** currently has pinned `electron ^1.0.0`. If we release `2.0.0`, 9.x can not install `2.0.0` with the default command `cordova platform add electron`... 

Users though can still install `2.0.0` by providing the version: `cordova platform add electron@2.x`.

www/docs/en/9.x/platform_pinning/index.md
janpio: I didn't suggest that you take this sentence, but asked for a better way to say what the sentence contains.

www/docs/en/9.x/platform_pinning/index.md
janpio: No. What does `7.1.1` actually mean with `will fetch the latest version up to a patch release.`? What version would it not use?

www/docs/en/9.x/platform_pinning/index.md
janpio: But why "highly not recommended"?

www/docs/en/9.x/platform_pinning/index.md
janpio: See https://github.com/apache/cordova-docs/pull/991#discussion_r284200014

www/docs/en/dev/platform_pinning/index.md
```diff
@@ -24,11 +24,11 @@ description: Cordova CLI pinned platform versions.
 
 ## Platform Pinning
 
-Cordova CLI pins the supported platform version in conjunction with the version of CLI that is being used.
+Cordova CLI manages a list of pinned Apache Cordova maintained platforms. When there is a major release of Cordova CLI, the platform's pinned version is updated to the latest major released version. Typically, the updating of this pin is made only on a major release of CLI. This is because the pinned versions are pinned with a caret `^` allowing CLI to continue to fetch new minor and patch releases of any given pinned platform.
```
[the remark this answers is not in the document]
erisu: I will leave that up to the developer to decide when they do the release.

In most cases when we release a major CLI we had already released a major of each platform. There is typically no `.1` or `.2` or what ever at that time unless someone really screwed up a release or introduce a major break/security bug before CLI was released. Then yes the developer can start on that version...

If for any reason a platform was not released, the developer could also decided at that time to update to the last release but I think it would look a bit ugly... Also people may feel at that point that past minors/patches are no longer consider supported when bumped. It more like a range of what works and currently supported with the know major CLI.

Since we have the `^` it will always use latest minor/patch.

Maybe it can be argued, if it works, to just drop the `^` and change the entire format to `8.x` then it is always major. But this is not something to discuss in a PR about what we have now for docs.

Anyways there is many up to developer/releaser information and use cases. All this can be documented else where like contribute. I think all this is not necessary for end user to read. This generally outlines that we pin, when pins are updated, what would potential fetched.

www/docs/en/dev/platform_pinning/index.md
```diff
@@ -43,8 +43,6 @@ Available platforms:
   windows ^7.0.0
 ```
 
-_Note: When a platform is installed, the "**Installed platforms:**" section will display the actual installed platform version. The installed platform will no longer display in the "**Available platforms:**" section until it is removed from the project._
-
-Starting from **Cordova CLI 9** and up, all platforms are pinned with an caret (`^`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a minor release.
+Using the above information, when `cordova platform add android` is executed, the latest minor/patch release version starting from 8.0.0 or after (8.x.x) will be fetched. If a version is provided, it will fetch the given version. E.g `cordova platform add ios@5.0.1` will fetch Cordova iOS 5.0.1.
 
-In **Cordova CLI 8** and below, all platforms are pinned with a tilde (`~`). It means that by installing a platform without a specified version, Cordova will fetch the latest version up to a patch release.
```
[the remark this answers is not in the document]
erisu: Cordova 8 should be in 8.x docs. This is dev that can be piped back into 9.x... There is no need t keep this here. Someone else can make a PR to update old release docs.

www/docs/en/9.x/platform_pinning/index.md
erisu: Outdated: Refactored

www/docs/en/9.x/platform_pinning/index.md
erisu: Outdated: Refactored

www/docs/en/9.x/platform_pinning/index.md
erisu: Outdated: Removed

www/docs/en/dev/platform_pinning/index.md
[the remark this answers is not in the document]
janpio: I think you misunderstood.

As example right now Cordova iOS is 5.0.1, pinned to `^5.0.0` in lib. If we do a iOS 5.1.0 now, and then release a new CLI, if I understood what you wrote correctly, there would be no need to update the pinning of Cordova iOS because the major didn't change.

But I think it would actually be beneficial to change the pinning to `^5.1.0` - even though it wouldn't change the installed version. But it improves the output you see when running `cordova platform list` and during  `cordova platform add ...` before it knows the actual version - it would give the user a better picture of what verison they will get.

www/docs/en/9.x/platform_pinning/index.md
erisu: Outdated: Refactored

www/docs/en/dev/platform_pinning/index.md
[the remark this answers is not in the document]
erisu: If we want to provide a `user a better picture of what version they will get.` then we should be fetching a list of versions from npm and then filter it by our pinned version and then grab the latest. Then it would show `5.0.1`, `5.1.0` or even `5.1.2`... 

Right now in the current structure, if we want to provide that information, we would have to do a patch release of LIB and CLI for every platform release. Break the release schedule down for this:

* Prepare and Submit Platform Release Vote 1 day
* 48 hours (2 days) release & prepare lib release vote
* 48 hours (2 days) release lib & prepare CLI release vote
* release
(Total time on current structure: approx. around 1 week)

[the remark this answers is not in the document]
erisu: Anyways, if there was a patch release to LIB and CLI, the developer can update the pinnings at that time if wanted.. There isn't anything blocking them from doing it when they want. Just that Major to Major is typical. Not a major platform in a minor or patch of lib/cli.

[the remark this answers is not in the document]
janpio: This is not what I suggested.

You define here that pinning only uses major versions:

> the platform's pinned version is updated to the latest major released version.

This is unnecessarily restricting. Use the latest or at least latest minor to pin, and the existing codebase can give more useful output to users. (Pre Cordova 9 we already pinned the latest available version of a platform - so I don't understand why this new limitation has to be added here just because the current pinning list only contains .0.0 versions.)

```suggestion
Cordova CLI manages a list of pinned Apache Cordova maintained platforms. When there is a major release of Cordova CLI, the platform's pinned version is updated to the latest released version. Typically, the updating of this pin is made only on a major release of CLI. This is because the pinned versions are pinned with a caret `^` allowing CLI to continue to fetch new minor and patch releases of any given pinned platform.
```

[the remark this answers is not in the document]
erisu: Going back to what I said before,

> When there is a major release of Cordova CLI, the platform's pinned version is updated to the latest major released version. 

Typically by the time we release Cordova CLI 10 or any major version, it typically also inlines with a major platform release.

```
  android ^9.0.0
  browser ^7.0.0
  electron ^2.0.0
  ios ^6.0.0
  osx ^6.0.0
  windows ^8.0.0
```

But I also said in the past comment 

> If for any reason a platform was not released, the developer could also decide at that time to update to the last release but I think it would look a bit ugly...

That means the developer can change it to `^5.1.0` or what ever version it currently at. I just don't document all possible developer choice cases.

I can change the wording to this

```
When there is a major release of Cordova CLI, the platform's pinned version is updated latest released version. 
```

But I know in almost 99% of the time its major.

PR #994, opened 2019-05-14
Improve ESLint setup and fix linting errors

raphinesse:
### Motivation and Context
- Not all JavaScript files were linted
- ESLint dependencies were outdated

### Description

- Lints all build-related JavaScript files
- Updates ESLint dependencies
- Fixes all resulting linting errors


tools/bin/fetch_docs.js
```diff
@@ -94,7 +87,7 @@ function getFetchedFileConfig (entry) {
 
     // complete src config
     if (!srcConfig.packageName) {
-        srcConfig.packageName = packageNameFromRepoName(srcConfig.repoName);
+        srcConfig.packageName = srcConfig.repoName.split('/')[1];
```
dblotsky: It is important for this transformation to have a name for readability. Why the removal?

dblotsky:
Nice change! Linting is awesome.

tools/bin/fetch_docs.js
raphinesse: The transformation function essentially became this one liner after removing unused variables.

Since we are setting a property named `packageName` to the value of a single expression involving only a property named `repoName` as input, I felt that a having a function named `packageNameFromRepoName` did not really improve clarity. IMHO, being able to see the _how_ right there where we use it adds more value in this case.

That being said, I don't particularly mind the function being there. So if you insist, I can bring it back of course.

dblotsky: You're correct: the function is very simple. Its purpose is to isolate the repo-to-package logic (however simple it may be) from the other logic in the function. If it's not too much trouble to bring it back, please do. That way this code can be easier to debug/change if/when it needs to be.

PR #1009, opened 2019-07-16
Android: App Bundle Documentation

breautek:
### Platforms affected
Android

### Motivation and Context
This PR is a brother PR of https://github.com/apache/cordova-android/pull/764 that adds documentation related to the new app bundle feature.

### Description
Documented the new acceptable key in `build.json` or via command line, `--packageType`, which accepts a value of `apk` or `bundle`. Defaults to `apk`.

### Testing
I ran `npm serve` to generate the website, and observed my changes locally.
I also ran `npm test`, which builds successfully with "Possible 404 in ToC" warnings. The warnings however appears in master and because they are warnings I did ignore them.

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/android/index.md
+++ b/www/docs/en/dev/guide/platforms/android/index.md
@@ -328,7 +328,7 @@ To sign an app, you need the following parameters:
 | Alias                 | `--alias`         | The id specifying the private key used for signing
 | Password              | `--password`      | Password for the private key specified
 | Type of the Keystore  | `--keystoreType`  | *Default: auto-detect based on file extension*<br>Either pkcs12 or jks
-| Package Type          | `--packageType`   | *Default: apk*<br>Builds an APK or a bundle (.aab) file.<br>Accepts either `apk` or `bundle`
+| Package Type          | `--packageType`   | *Default: apk*<br>Specify whether to build an APK or an [Android App Bundle] (https://developer.android.com/guide/app-bundle) (.aab) file.<br>Accepts either `apk` or `bundle`
 
 These parameters can be specified using the command line arguments above to
 the [Cordova CLI][cli_reference] `build` or `run` commands.
--- a/www/docs/en/dev/guide/platforms/android/index.md
+++ b/www/docs/en/dev/guide/platforms/android/index.md
@@ -328,13 +328,14 @@ To sign an app, you need the following parameters:
 | Alias                 | `--alias`         | The id specifying the private key used for signing
 | Password              | `--password`      | Password for the private key specified
 | Type of the Keystore  | `--keystoreType`  | *Default: auto-detect based on file extension*<br>Either pkcs12 or jks
+| Package Type          | `--packageType`   | *Default: apk*<br>Builds an APK or a bundle (.aab) file.<br>Accepts either `apk` or `bundle`
 
 These parameters can be specified using the command line arguments above to
 the [Cordova CLI][cli_reference] `build` or `run` commands.
 
 __Note__: You should use double `--` to indicate that these are platform-specific arguments, for example:
 
-`cordova run android --release -- --keystore=../my-release-key.keystore --storePassword=password --alias=alias_name --password=password`.
+`cordova run android --release -- --keystore=../my-release-key.keystore --storePassword=password --alias=alias_name --password=password --packageType=bundle`.
 
 ### Using build.json
 
@@ -350,14 +351,16 @@ build configuration file:
             "storePassword": "android",
             "alias": "mykey1",
             "password" : "password",
-            "keystoreType": ""
+            "keystoreType": "",
+            "packageType": "apk"
         },
         "release": {
             "keystore": "../android.keystore",
             "storePassword": "",
             "alias": "mykey2",
             "password" : "password",
-            "keystoreType": ""
+            "keystoreType": "",
+            "packageType": "bundle"
         }
     }
 }


www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -328,13 +328,14 @@ To sign an app, you need the following parameters:
 | Alias                 | `--alias`         | The id specifying the private key used for signing
 | Password              | `--password`      | Password for the private key specified
 | Type of the Keystore  | `--keystoreType`  | *Default: auto-detect based on file extension*<br>Either pkcs12 or jks
+| Package Type          | `--packageType`   | *Default: apk*<br>Builds an APK or a bundle (.aab) file.<br>Accepts either `apk` or `bundle`
```
janpio:
```suggestion
| Package Type          | `--packageType`   | *Default: apk*<br>Wether to build an APK or a Android App Bundle (`.aab`) file.<br>Accepts either `apk` or `bundle`
```

janpio:
This PR shows that this being buried under "Signing an App" is probably not enough. Can we somehow make this more prominent? Maybe similar to https://cordova.apache.org/docs/en/latest/guide/platforms/ios/index.html#xcode-build-flags?

A link to https://developer.android.com/guide/app-bundle on the first mention of "App Bundle" would probably also be nice so that users can get some more information about it.

breautek:
> This PR shows that this being buried under Signing an App`is probably not enough.

I agree @janpio  I can mimic the iOS docs here. It looks like mostly an addition, however they do lack `--buildFlag` in the table under Using Flags... but they do have `--packageType` in the table. Perhaps in android docs, we will also include `--packageType` in the Using Flags table, but go in a bit more detail on  using `apk` or `bundle` below? Including the link to the android guide as you suggested (or should the link still be on the first mention?)

janpio:
Feel free to optimize both tables and documentation in general. 

The crux behind the differences is that on iOS the `packageType` indeed is a codesigning thing, while on Android it is a "build type" thing. So maybe we have to think again if this really was a good idea to reuse that parameter or if we are OK with that discrepancy.

breautek:
It does feels right to have it in the `build.json`. I think it's actually nice that you can specify `apk` package type for debug, but have `bundle` package type for release. But the documentation structure makes it weird, perhaps there should be a section for building an app that mirrors the signing the app section.

For example:

## building the app
...
### Using build.json
[shows table flags related to build options]

## signing the app
...
### Using build.json
[shows table flags related to signing options...]

Then of course all my changes will be moved under the building the app section.

janpio:
That might be a good idea - `build.json` and `--buildConfig` is a concept that applies to all platforms, and it seems it can be used for multiple use cases. https://cordova.apache.org/docs/en/latest/guide/platforms/ios/index.html#xcode-build-flags could also be moved into the "build" section then. We would need a more general documentation for `build.json` and `--buildConfig` then - which is useful anyway. 

But maybe we should move that specific part to a new issue and PR to not add to much in here?

---

> I think it's actually nice that you can specify `apk` package type for debug, but have `bundle` package type for release.

I agree with _that_, but my concern was specific to `packageType` having two different positions here:

> on iOS the `packageType` indeed is a codesigning thing, while on Android it is a "build type" thing.

breautek:
> on iOS the packageType indeed is a codesigning thing, while on Android it is a "build type" thing.

Are you sure on that? The docs say...

> This will determine what type of build is generated by Xcode. Valid options are development (the default), enterprise, ad-hoc, and app-store.

And based on the [Apple docs](https://developer.apple.com/library/archive/documentation/ToolsLanguages/Conceptual/DevPortalGuide/DistributinganApp/DistributinganApp.html)

It just sounds like different distribution formats. Which is essentially the same as the android APK and bundle format. I don't think on iOS `packageType` is a codesigning thing either. Regardless of the chosen `packageType` they all need to be signed for distribution, and I think that applies to iOS as well.

janpio:
That is not false, but also not strictly correct at the same time: For iOS the `packageType` defines what kind of signing credentials are used. That also decides what kind of end result you get (similar to APK vs. Bundle), but they are all `.ipa` files - just signed in different ways - so in this case I feel it is more limited to signing. The build that is signed is identical.

www/docs/en/dev/guide/platforms/android/index.md
brodybits: Suggestion by @janpio, with an added word (maybe not needed), spelling fix, and a -> an (spoken language):

```suggestion
| Package Type          | `--packageType`   | *Default: apk*<br>Specify whether to build an APK or an Android App Bundle (`.aab`) file.<br>Accepts either `apk` or `bundle`
```

www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -350,14 +351,16 @@ build configuration file:
             "storePassword": "android",
             "alias": "mykey1",
             "password" : "password",
-            "keystoreType": ""
+            "keystoreType": "",
```
brodybits: Too bad we have to touch multiple lines whenever we add a new entry to the end. I guess not so much that we can do right now:(

breautek:
> I am approving these changes on the condition that [apache/cordova-android#764](https://github.com/apache/cordova-android/pull/764) is finished and merged first. Sorry for the extra noisy reviews.

I am in progress of making changes outlined back at https://github.com/apache/cordova-docs/pull/1009#issuecomment-512227371 so there will be incoming PR changes. Or are you suggesting to keep what I have currently and leave the re-organizing to another PR?

brodybits:
> I am in progress of making changes outlined back at [#1009 (comment)](https://github.com/apache/cordova-docs/pull/1009#issuecomment-512227371) so there will be incoming PR changes.

I would be happy to review any further changes that you feel are needed.

> Or are you suggesting to keep what I have currently and leave the re-organizing to another PR?

My vote would be to add the required info to document how the feature can be enabled from the command line or enabled by configuration. Just enough information so that someone already familiar with this AAB functionality knows how to use it.

So yes, I would favor that we “leave the re-organizing to another PR”.

Thanks again for your hard work on this functionality!

www/docs/en/dev/guide/platforms/android/index.md
janpio: I mentioned it before: Link "Android App Bundle" to https://developer.android.com/guide/app-bundle for context.

breautek: I have these suggested changes in a pending commit, including the Android App Bundle link

breautek:
> Feel free to optimize both tables and documentation in general.
> 
> The crux behind the differences is that on iOS the `packageType` indeed is a codesigning thing, while on Android it is a "build type" thing. So maybe we have to think again if this really was a good idea to reuse that parameter or if we are OK with that discrepancy. (ping @dpogue)

So at this point I believe this is the last major thing preventing this from moving forward. I'll add my thoughts on this...

Even though `packageType` is a code-signing thing for iOS, I think the metaphor still fits. It is still choosing a package or deployment strategy essentially. Yes, each `packageType` will all yield an `ipa` file, just signed differently. And depending on how the `ipa` file is packaged... depicts on how you can use that `ipa` file.

Likewise for android, following the current naming convention, they have 2 `packageType` options, `apk` and `bundle`. While they may produce two different files, in the end, they are still a package for deployment, just like iOS. The package could be deployed locally to the phone, or it could be uploaded to the play store. So just like iOS, I think the `packageType` metaphor still fits.

janpio:
I would be ok with that. 

To gather more feedback, you should probably send an email to the dev@ mailing list. First explain the PR in general, then mention that this is a problem that concerns were voiced about. Explain it neutrally, then add your explanation. If there are no -1 or objections, we can go ahead and merge this.

breautek:
Pinging the reviewers as requested from the mailing list discussion

@dpogue
@erisu 
@brodybits 
@janpio

PR #1010, opened 2019-07-17
docs(ios): Improve "Using Flags" and "Using build.json"

janpio:
### Platforms affected

iOS

### Motivation and Context

When reading up on the usage of `build.json` I found some missing information for iOS.

### Description

- Add information what commands these flags can be used on
- Add example of their usage
- Describe how `build.json` is actually used: by default, or via other file name as param buildConfig
- Add info about mix and match (via Android)

#### Open questions

- Is "**Note**: You should use double `--` to indicate that these are platform-specific arguments, for example:" actually correct?
- Is "There is also support to mix and match command line arguments and parameters in build.json. Values from the command line arguments will get precedence. " actually true?

---

If this PR is merged, we should probably check if some of the changes can/should also be applied to Android at https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html


janpio:
> * Is "**Note**: You should use double `--` to indicate that these are platform-specific arguments, for example:" actually correct?

Per Slack @erisu mentioned that this is not correct for iOS. I should confirm this, then investigate where this difference comes from (cordova-ios vs. cordova-android). Afterwards this PR should be adapted (and possibly some changes to Android done as well).

erisu (approved):
LGTM

PR #1011, opened 2019-07-24
iOS support for IndexedDB

jgaehring:
iOS has supported IndexedDB since the release of iOS Safari 10.

### Platforms affected
iOS

### Motivation and Context

Support for IDB has long been available on iOS. My team recently migrated our db from WebSQL to IDB and are enjoying full support on the iOS platform. We would have gone with IDB from the start but the docs led us to believe there wasn't support. Also, WebSQL is deprecated and IDB is now the best cross-platform solution available, imo. Hopefully this saves some other developers grief.

### Description

Simply adding iOS to the list of supported platforms, and removing it from the list of disadvantages.

### Testing

We have IDB in production on iOS and it is working just fine.

### Checklist

### DIFF
--- a/www/docs/en/dev/cordova/storage/storage.md
+++ b/www/docs/en/dev/cordova/storage/storage.md
@@ -166,6 +166,7 @@ IndexedDB is supported by the underlying WebView on the following Cordova platfo
 
 - Windows (with some limitations)
 - Android (4.4 and above)
+- iOS (with Safari 10 and above)
 
 ### Windows Limitations
 
@@ -268,7 +269,6 @@ For more information, see:
 
 ### Disadvantages
 
-- Not supported on iOS.
 - Complex API with nested callbacks.
 - Limited total amount of storage (typically around 5MB).
 


brodybits:
Thanks @jgaehring for the contribution, please accept our apologies for the slow review. I have just sent a request for review from a cordova-ios specialist.

In case you do not hear anything in the next few weeks, please ping us again and I would be happy to check the information.

Please feel free to contact us by Slack or on the mailing list, just follow the appropriate link from the footer of cordova.io or cordova.apache.org.

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -166,6 +166,7 @@ IndexedDB is supported by the underlying WebView on the following Cordova platfo
 
 - Windows (with some limitations)
 - Android (4.4 and above)
+- iOS (with Safari 10 and above)
```
jcesarmobile:
```suggestion
- iOS 10 and above
```
I would change the iOS line to something like this, because

1. People might not know the safari version matches the iOS version
2. Sounds like it's only supported in Safari, but it's also supported on the WebView and apps that use WebView (Cordova apps, Chrome, etc)

jcesarmobile:
I've double checked and it was really supported on iOS 8 on Safari and WKWebView, but was very buggy.
On iOS 10 the bugs are supposedly fixed, and was included on UIWebView too.

But I don't think we need to give all those details, and since current cordova-ios only supports iOS 10+, maybe we can just say iOS and don't specify a version?

brodybits:
> maybe we can just say iOS and don't specify a version?

Agreed on my part.

brodybits:
I would like to close this in favor of a more comprehensive update that I proposed in PR #1042. I am giving co-author credit to @jgaehring for the proposal in PR #1042.

jgaehring:
Great! Glad to see the updates are moving forward. Do feel free to close this.

brodybits:
Closing in favor of PR #1042, as agreed with @jgaehring - thanks again!

Issue #995, opened 2019-05-15
Remove all translations


raphinesse:
Doc translations are not being actively maintained and are thus outdated. Furthermore, at least some of the existing translations seem to be really bad too. Consequently I propose to remove all translations from the docs.

I checked [the guide entry page](https://cordova.apache.org/docs/en/latest/guide/cli/index.html) in German and it's so horribly bad that it can be considered totally useless IMO. OTOH, doing a Google translation of the English version yields a really helpful and comprehensible text.

Related to https://github.com/apache/cordova/issues/29

erisu:
I am also in favor of removing the translated docs. 👍

dblotsky:
Many people use old versions of Cordova, and the docs for those versions are still valid.

Also, why not dedicate effort to reviving translations?

raphinesse:
@dblotsky

We have very _limited resources_. We should _not_ put _any_ effort into supporting old versions of Cordova let alone 10 translations of their docs.

But they do generate work by just being there: longer build times, problems during updates of dependencies, increased complexity of build, increased repository checkout size, ...

Also, as I already stated above, the quality of at least done of these translations is so bad, that no value can be attributed to them whatsoever.

janpio:
> Also, why not dedicate effort to reviving translations?

Nobody was interested the last time we had this discussion, nobody popped up in the meantime, and even if somebody appears now they have to start from 0 as adapting what _is_ there is probably more work than just starting from nothing.

---

In my previous suggestions on this topic, I normally only thought about removing translations from a future/current CLI version. But in the last few months, my opinion changed as well to us removing the existing translations completely. Even for old versions they are broken and misleading.

Only thing we have to take care of is that all the old links redirect to their correct English versions. This is a challenge, as sometimes there exist (outdated) translations for content that does not exist in the English version of the docs - so we will need to spend some effort to figure that out and find good redirection targets.

dblotsky:
I think I agree then: if we have no resources to maintain translations, it's better to remove them than have bad ones.

szotp:
Please remove translations ASAP, my search results for cordova are polluted with these machine translated, indecipherable titles. I had to change my browser's language settings just to be able to do my tasks.

Issue #841, opened 2018-07-07
About Chinese Simplified translation


Ray-Eldath:
**Issue Type**
[ ] Bug
[X] Feature Request

**Priority**
[ ] Minor 
[X] Major 
[ ] Critical 
[ ] Blocker 

**Description**

The Simplified Chinese [简体中文] translation of cordova-doc is **totally incorrect**! Actually, it's *just a copy* of  Traditional Chinese translation.
I think you can use Traditional Chinese - Simplified Chinese converter which is being used by Wikipedia. See: [Automatic conversion between simplified and traditional Chinese](https://meta.wikimedia.org/wiki/Automatic_conversion_between_simplified_and_traditional_Chinese) and [MediaWiki / LanguageZh.php](https://github.com/wikimedia/mediawiki/blob/05c72943a7bb06f618c7bd6f01cbbc2ce0665178/languages/classes/LanguageZh.php)
But I think there is high-quality Simplified Chinese translation already, maybe you can use theirs. See: [Cordova中文网](http://cordova.axuer.com)

jcesarmobile:
Thanks for the issue, but translations are going to be removed, so this won’t be fixed

PR #759, opened 2017-11-13
[Android] Chinese translation

kungfucode-rex:
### Platforms affected
Android platform

### What does this PR do?
Doc translation of Chinese for Android platform

### What testing has been done on this change?
none

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #742, opened 2017-09-29
update Chinese translation on 7.x/guide/platforms/ios/config.md

situee:
fixed some Chinese (zh-cn & zh-tw) translation errors

### Platforms affected
/7.x/guide/platforms/ios/config.md

### What does this PR do?
fixed some Chinese translation errors

### What testing has been done on this change?
checked on preview

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #1015, opened 2019-09-05
chore: remove translation

erisu:
### Motivation and Context

See #995

### Description

- Removes Translated Docs

closes: #995
closes: #932
closes: #841
closes: #759
closes: #742
closes: #735

### Checklist

- [x] I've deleted documentation that is necessary


dblotsky:
The ask in the issue was to remove translations, but not versions. Do we have proof that nobody uses Cordova < 5.4.0?

Dmitry

janpio:
What happens to the old URLs - are they redirected to `latest`? Error message with proper 404 or 410?

erisu:
@dblotsky 

Sadly, telemetry did not exist in 5.4.0 or lower so I can not tell you for certain whats the % of users that uses 5.4.0 or less. Here is an example of what I see from the data we do have.

**New Users (Sep 5th, 2019)**
5.x => ???
6.5.0 => 1,673 (2%)
7.x => 8,065 (10%)
8.x => 21059 (25%)
9.x => 53,069 (63%)


**Users out of 17,406 (Sep 3rd, 2019)**
5.x => ???
6.5.0 => 2.06%
7.x => 7.63%
8.x => 32.54%
9.x => 57.22%

If there was users still using 5.x and we throw that into the mix, I am guess it would be 1% or less.

erisu:
@dblotsky 

In the original ticket, after you raised up the concern, one of @raphinesse replies contained:

> We should not put any effort into supporting old versions of Cordova let alone 10 translations of their docs.

I kinda took this into account and also removed versions but kept only a few based on the telemetry data. 

If other are as well wanting to keep all versions or add back a few more versions back, I can update. This is why I did the deletion of versions as a separate commit in this PR to easily delete or update if necessary.

erisu:
Here is also the telemetry link for reference: http://telemetry.cordova.io

raphinesse:
Well, then let's just remove the translations first. We can figure out what we do about those old versions later.

Thanks for taking care of this, @erisu.

erisu:
Redirects to 404

dblotsky:
Redirecting to 404 would be a pretty awful experience.

There is a bit of process for obsoleting pages: https://github.com/apache/cordova-docs/blob/master/doc/redirects.md.

But, I think obsoleting all translations is a big enough deal that we should:
1. Announce it (e.g. in a blog post),
2. Come up with a new redirect page specific for language deprecation (preferably in that language).
3. Have a roadmap for adjusting the build system not to build for all those languages.

janpio:
> Come up with a new redirect page, specifically for language deprecation.

Here it would probably make sense to redirect to the EN version of everything.

dblotsky:
We have a deprecation page in many languages. We can use those at the very least. But, what do you guys think, from a user experience perspective, of a custom message in each language explaining why we decided to remove translations?

raphinesse:
Regarding your requested for changes @dblotsky:

> Redirecting to 404 would be a pretty awful experience.

Agreed.

> I think obsoleting all translations is a big enough deal that we should:
> 1. Announce it (e.g. in a blog post),

That would definitely be great to have.

> 2. Come up with a new redirect page, specifically for language deprecation.

Here I would go with what @janpio proposed:

> Here it would probably make sense to redirect to the EN version of everything.

I think this is more helpful than a redirect to a deprecation message. And it's very easy to implement. The following snippet added to [`www/.htaccess`](https://github.com/apache/cordova-docs/blob/d0890966b46655bbdeb8f2bf980b09a1f5bef525/www/.htaccess#L75) _should_ do the job (_untested_):

```apache
# Redirect all docs translations to their English version
RewriteCond $1 !=en
RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/(.*)$ {{site.baseurl}}/docs/en/$2 [R=301,L]
```

> 3. Have a roadmap to change the build system not to build for all those languages.

IMHO, that is a matter of creating an issue titled _Remove language support from build system_ but should not block us from going live with these changes. It can be addressed when someone finds the time to do so.

brodybits:
Any suggestions how we can move this forward?

Anything that I can do?

dblotsky:
@brodybits sorry for the delay, and thank you for patience!

I think the best way forward is the one that's easiest to review: so, as few changes at a time as possible.

I propose this plan (possibly as 4 sequential PRs, with a week between each):
1. Announcing the deprecation in a blog post,
2. Adding the redirects, deprecation pages, etc.,
3. Actually deleting the translated pages (since the redirects made them unreachable), and
4. Deleting the code that supports translations.

What do you think?

erisu:
Closing, build or deploy performance at this time is not a priority.

PR #1020, opened 2019-10-10
[BUG] Clarify sponsorship in the footer

brodybits:
### Platforms affected

All

### Motivation and Context

Existing sponsorship link which shows up on every page in our documentation goes to an ASF "sponsorship" page but it is not so clear that the sponsorship would go to non-coding activities. This is not in line with our needs for project sustainability (apache/cordova#163).

I think we should also direct people to the actual project sustainability issue (apache/cordova#163).

I am including "[BUG]" in the description since I think this is something that needs to be fixed in our documentation.

### Description

- Add a footnote to the Apache "sponsorship" link to *clearly* show the following:
  - that the existing "sponsorship" would go to non-coding activities
  - link to the project sustainability issue (apache/cordova#163)

### Testing

TODO:

- [ ] check that generated documentation renders correctly

### Checklist

- ~~I've run the tests to see all new and existing tests pass~~
- ~~I added automated test coverage as appropriate for this change~~
- ~~Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)~~
- ~~If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))~~

### DIFF
--- a/www/_includes/footer_contents.html
+++ b/www/_includes/footer_contents.html
@@ -37,6 +37,7 @@
                     </li>
                     <li>
                         <a target="_blank" href="http://www.apache.org/foundation/sponsorship.html">Become a Sponsor</a>
+                        <sup><a href="sponsorship">*</a></sup>
                     </li>
                     <li>
                         <a target="_blank" href="http://www.apache.org/foundation/thanks.html">Thanks</a>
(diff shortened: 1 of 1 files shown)


brodybits:
> Inappropriate.

What?!?!?

brodybits:
Closing this one out for now. I would like to say a couple things for the record:

I continue to stand by all of the points I made in the motivation. I think this continues to be a pretty bad problem, if the "sponsorship" link cannot cover one of the most important pillars.

Some of the feedback seems to be a bit subjective.

I would definitely be happy to leave this up for discussion, if there is any way we can find a solution to the problem.

jcesarmobile:
We have been offered one sponsorship in 10 years, so I don't think this clarification is needed. Whoever wants to donate to Apache should read their sponsorship policy first, and if they want to fund development work should offer it directly to some developer.

Issue #1035, opened 2019-12-12
Automatic doc build is broken (Travis CI)


brodybits:
# Bug Report

### What is expected to happen?

UPDATED: Automatic doc build should be done whenever we add a release note or update any other part of the documentation.

Travis CI build should just keep working whenever we merge PRs such as #1032.

### What does actually happen?

Travis CI build stopped working right after I merged PR #1032.

_Automatic doc builds from release notes and other updates is broken._

It looks to me like there is an issue with svn.

I am not an Apache website svn build expert. I really hope we can get the right expertise to look at this one.

### Command or Code

see .travis.yml

### Environment, Platform, Device

see .travis.yml

### Version information

N/A

## Checklist

- [x] I searched for existing GitHub issues
- ~~I updated all Cordova tooling to most recent version~~
- [x] I included all the necessary information above

timbru31:
Ping @janpio - I remember there were some issues with docs in the past related to SVN and credentials.

timbru31:
CI failure notice:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

janpio:
The SVN stuff is using my credentials, possible that they got blocked. Do not have a setup to confirm, so you might just want to replace with yours. (There is no real other solution possible as they this is pushed onto SVN is not supported by ASF Infra)

brodybits:
I raised https://issues.apache.org/jira/browse/INFRA-19575 to get some help from INFRA.

janpio:
This has been raised before (a fews times actually), and there are multiple other solutions for websites in ASF - non which match the setup Cordova is currently using.

brodybits:
Thanks @janpio. As I said in INFRA-19575, I would really favor that we find a way to work with Apache INFRA, rather than using individual dev credentials from myself or anyone else.

brodybits:
This issue affects automatic build of release notes such as cordova-node-xcode@2.1.0 (PR #1037). I just updated the title and description, pinning now.

NiklasMerz:
Is the automatic build working again? I was about to do a manual deploy but master was already deployed with a SVN commit by janpio.

erisu:
Closing ticket as the automatic build is working.

PR #1038, opened 2019-12-14
build: major update to jekyll 3.8

raphinesse:
Fixes after update:
- Fix access to layout variables

Changes in generated site:
- Post dates are converted to build-local TZ
- Irrelevant whitespace changes
- Irrelevant changes to `sitemap.xml` (ordering of entries)

While Jekyll 4 is out already, it necessitates switching from `redcarpet` to `kramdown`. Unfortunately, that causes so many changes in the generated markup that its correctness cannot be verified by simple diffing anymore.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
-gem 'jekyll', '= 2.5.3'
+gem 'jekyll', '~> 3.8'
 gem 'redcarpet', '= 3.3.2'
 gem 'rouge', '= 2.1.0'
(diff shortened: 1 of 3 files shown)


brodybits:
Thanks @raphinesse. I would *personally* be very reluctant to make this kind of an update until we fix the website build (see #1035).

raphinesse:
@brodybits Agreed. I did not realize that the build was currently broken. I'm in no hurry with this PR.

Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
-gem 'jekyll', '= 2.5.3'
+gem 'jekyll', '~> 3.8'
```
dblotsky: Why a `~` version and not `=`?

raphinesse: To allow non-major updates that are not expected to introduce breaking changes.

This is in line with our JavaScript dependency management where we use npm's default, i.e. `^` version ranges. If there are good reasons to prefer version pinning that are specific to RubyGems, we can revert to using a pinned version of course.

breautek:
I wonder if we could revive this PR -- our current dependencies is becoming quite obsolete I fear it's going to start getting difficult setting up an actual working environment capable of building docs, especially for newcomers.

breautek (approved):
lgtm

erisu (approved):

NiklasMerz (approved):

PR #1042, opened 2019-12-30
Update IndexedDB platform support

brodybits:
### General

This proposal supersedes and closes #1011 (PR #1011)

/cc @jgaehring with thanks for the original proposal in #1011

### Platforms affected

Android, browser, Electron, iOS, "osx" (macOS)

### Motivation and Context

From a second look at PR #1011 I discovered that the documentation of IndexedDB support is outdated for multiple platforms. I would like to get the documentation updated for all affected platforms at once.

### Description

- Update the description to make it clear that this functionality is supported by the underlying WebView on **all** platforms, with possible limitations added for the (web) browser platform
- The "4.4 and above" condition for Android now gone, since we have already dropped support for Android pre-4.4
- "Not supported on iOS" item now completely and unconditionally gone from disadvantages section due to this comment: https://github.com/apache/cordova-docs/pull/1011#issuecomment-567250310

### Testing

Visual inspection

### Checklist

- [ ] [TODO] I've run the tests to see all new and existing tests pass
- ~~I added automated test coverage as appropriate for this change~~
- ~~Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)~~
- [x] If this Pull Request resolves ~~an issue~~ *another PR*, I linked to the ~~issue~~ *PR* in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation ~~if~~ *as* necessary

FOLLOWUP TODO:

- [ ] If merged, we may need to do a manual rebuild (see #1035)


www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -162,10 +162,12 @@ IndexedDB provides a simple and easy to understand data model, much like LocalSt
 But unlike LocalStorage, you can create multiple databases, with multiple stores per
 database, and its asynchronous API and search indexes provide performance benefits.
 
-IndexedDB is supported by the underlying WebView on the following Cordova platforms:
+IndexedDB is supported by the underlying WebView on all platforms,
+with known limitations on web browser and Windows.
```
timbru31: web browser should be plural, shouldn't it?

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -162,10 +162,12 @@ IndexedDB provides a simple and easy to understand data model, much like LocalSt
...
 
-- Windows (with some limitations)
-- Android (4.4 and above)
+### Web browser limitations
+
+The actual behavior may depend on which browser is used. There may be differences on Edge and Firefox browsers, for example.
```
timbru31: IMO we should not use Edge as an example, because it soon is Chromium based, too. Maybe just use Safari or Chrome instead?

www/docs/en/dev/cordova/storage/storage.md
brodybits: I am reluctant to make "browser" plural since I am referring to the Cordova `browser` platform and would rather refer to it by proper name, if possible.

brodybits: What about this idea:

```suggestion
with known limitations on the following platforms:

- browser
- Windows
```

Or do you have any other ideas?

www/docs/en/dev/cordova/storage/storage.md
brodybits: How about this:

```suggestion
The actual behavior may depend on which browser is used. There could be differences between the behavior on Safari and Firefox browsers, for example.
```

vdshb:
Is 5 MB limitation correct for indexedDB at iOS, Android and browser? I'm new to this topic, but I thought indexedDB was invented to store more. Is it some Cordova plugin's additional limitation?

timbru31:
Good catch, no 5MB is not accurate anymore.  
For recent iOS/Safari version AFAIK there are no real sources and Android's limitation are here: https://stackoverflow.com/a/55483549/1902598 (the most up to date source, I've found).

breautek:
Chrome doesn't really give a limit in a traditional sense, but it tries to not consume all of the disk space. It does this by determining how much remaining space is available and will start implementing quotas if remaining space gets low.

Chrome will either try to keep at least 10% of the total storage or 2GB, whichever is lower, of the disk free, and will start implementing quotas once it reaches this threshold.

The table below comes directly from the chromium sources, and is accurately as of Chrome 76.
```
// The amount of the device's storage the browser attempts to
--
  | // keep free. If there is less than this amount of storage free
  | // on the device, Chrome will grant 0 quota to origins.
  | //
  | // Prior to M66, this was 10% of total storage instead of a fixed value on
  | // all devices. Now the minimum of a fixed value (2GB) and 10% is used to
  | // limit the reserve on devices with plenty of storage, but scale down for
  | // devices with extremely limited storage.
  | // *   1TB storage -- min(100GB,2GB) = 2GB
  | // * 500GB storage -- min(50GB,2GB) = 2GB
  | // *  64GB storage -- min(6GB,2GB) = 2GB
  | // *  16GB storage -- min(1.6GB,2GB) = 1.6GB
  | // *   8GB storage -- min(800MB,2GB) = 800MB
```

https://chromium.googlesource.com/chromium/src.git/+/refs/tags/76.0.3775.5/storage/browser/quota/quota_settings.cc

Not sure about iOS/Safari.

timbru31:
@dfahlander - sorry for the mention, but do you happen to know recent iOS and/or Safari quotas?

dpogue:
It's worth noting that the behaviour in web **browsers** can also differ from the behaviour in web **views**.

dfahlander:
@timbru31 I'm sorry but I don't know the quotas for iOS/Safari.

Aarbel:
@dfahlander @brodybits @timbru31 did you have more infos about indexedDb allowed memory size on iOS, and also on Android ? Thanks a lot !

breautek:
@aarbel

On android using the chrome browser, you can refer to this table https://github.com/apache/cordova-docs/pull/1042#issuecomment-611551394

It isn't clear on the quotas for wkwebview for ios. It isn't documented and afaik wkwebview isn't open sourced.

timbru31:
No, Safari is, at least to me, a blackbox. Last person to ask/tag is @beidson from Apple: Do you know any current size limits of the IndexedDB in iOS?

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -268,7 +275,6 @@ For more information, see:
 
 ### Disadvantages
 
-- Not supported on iOS.
 - Complex API with nested callbacks.
 - Limited total amount of storage (typically around 5MB).
```
brodybits, line 279: How about this resolution:

```suggestion
- Total amount of storage may be limited, depending on the platform and web view implementation.
```

breautek: Looks good to me.

erisu:
After PR #1043 is merged in, this PR should be rebased.

The only other minor feedback I have is that we do not need to break up the lines.

E.g.

> IndexedDB is supported by the underlying WebView on all platforms,
> with known limitations on the following platforms:

This could be written on one line.

> IndexedDB is supported by the underlying WebView on all platforms, with known limitations on the following platforms:

IMO, I look at markdown as a text document. When I write a technical specification document in Word, I usual do not do something like this.

But this is just my opinion and not a blocker.

timbru31:
Brody, could you rebase this PR so we can merge this in?

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -162,10 +162,17 @@ IndexedDB provides a simple and easy to understand data model, much like LocalSt
 But unlike LocalStorage, you can create multiple databases, with multiple stores per
 database, and its asynchronous API and search indexes provide performance benefits.
 
-IndexedDB is supported by the underlying WebView on the following Cordova platforms:
+IndexedDB is supported by the underlying WebView on all platforms,
+with known limitations on the following platforms:
```
timbru31, line 166:
```suggestion
IndexedDB is supported by the underlying WebView on all platforms, with known limitations on the following platforms:
```

timbru31:
Travis did not like the rebase and rapid commits. Once https://travis-ci.org/github/apache/cordova-docs/builds/741877609 finishes, I'll merge this in.

brodybits:
Thanks Tim for taking on the edit and merging!

PR #1062, opened 2020-02-22
doc(CONTRIBUTING): Remove text about JIRA and ICLA

NiklasMerz:
...add .vscode for users to avoid checking that in

### Platforms affected
Contributing docs

### Motivation and Context

This PR removes old information from the Contributing doc. AFAIK an ICLA is not required for trivial contributions, JIRA has been discontinued for Cordova. @purplecabbage Correct ?

### Testing

npm test

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ ruby_modules
 Gemfile.lock
 
 .*.swp
+.vscode/settings.json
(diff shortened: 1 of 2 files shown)


purplecabbage:
I would keep the iCLA text, but mention it is ‘recommended, only required for more substantial contributions’
Instead of removing jira issue link why not recommend they create a gh-issue in the relevant repo?
Jira is still required for security issues as we cannot make private gh-issues

NiklasMerz:
@purplecabbage Thanks for pointing that out. I was just checking how to do some things in docs and this file did not help or make any sense to me at the moment.

I just checkend and all repos use the exact same file. So changing this should be something global. Closing this now.

PR #1067, opened 2020-03-20
Adding Android SDK Tools info

whizkee:
In newer versions of Android Studio, you need to explicitly install the Obsolete Android SDK's `tools` before you can set the path variable.

### Platforms affected
All

### Motivation and Context
When following the current docs, you are instructed to set a path variable for android sdk tools.
If you are doing a fresh install of the latest android studio, you will not have the android sdk tools directory. 

Closes #1065 

### Description
Added short instructions about how to show obsolete packages and install android sdk tools.

### Testing
- Performed a fresh install of Android Studio 3.6.1; the tools directory was not created
- Installed Android SDK Tools after revealing obsolete packages ; the directory was created

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [xx] I've updated the documentation if necessary

PR #1072, opened 2020-04-03
Android: SDK tools

bkmgit:
Update documentation as per #296 since the tools may not appear as they have been deprecated and it appears that using command line tools is not fully supported yet.

### Platforms affected

Android

### Motivation and Context

Fixes documentation issue for Android SDK tools installation referred to in #926

### Description

Add information to check on show deprecated

### Testing

Went through installation

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/android/index.md
+++ b/www/docs/en/dev/guide/platforms/android/index.md
@@ -96,7 +96,7 @@ and make sure the following are installed:
 1. Android Support Repository (found under the "SDK Tools" tab)
 
 See Android's documentation on [Installing SDK Packages](https://developer.android.com/studio/intro/update)
-for more details.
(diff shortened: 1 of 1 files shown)


NiklasMerz:
Thank you for your contribution. Are you sure you linked the correct issues/pull requests? I cannot see information about your change in those since they are not related to Android.

But your change seems ok.

bkmgit:
Sorry should be https://github.com/apache/cordova-android/issues/926

NiklasMerz:
@bkmgit Thanks that makes this clearer

We also have #1067 . I think that one is to prefer.

bkmgit:
#1067 is nicely done. It would be helpful to merge it to save some time when people are installing. At the moment without the obsolete tools install, my Android build does not work. Tool chain may need to support both methods for a while. Docs will probably also need to be updated to use cmdline-tools as well.

erisu:
Resolved by #1067

PR #1079, opened 2020-04-22
A little change based on the system.

sajt:
### Checklist

### DIFF
--- a/www/docs/en/9.x/guide/cli/index.md
+++ b/www/docs/en/9.x/guide/cli/index.md
@@ -212,7 +212,7 @@ npm and you can search for them on the [plugin search page](/plugins/). Some key
 $ cordova plugin search camera
 ```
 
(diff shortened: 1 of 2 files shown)


breautek:
I approve but without the explicit green check because I think only the files inside `/dev/` is suppose to be modified. I'm pretty sure the other versioned files are suppose to be generated by a release script. 

Another member will confirm or correct me if I'm wrong.

erisu:
@breautek There is a script that can be used to update the current latest docs (9.x) from dev, but it will pull all of dev docs into 9.x.

```gulp snapshot```

If 10.x documentation is already pushed into dev, running the `snapshot` command might not be useful.

It might be OK to accept the direct change to 9.x.

breautek:
Thank you kindly for your contribution :+1:

PR #1008, opened 2019-07-15
Link to issue for Visual Studio 2019 on Windows platform

brodybits:

PR #1111, opened 2020-08-07
doc(windows): refactor requirements

erisu:
### Motivation and Context

- Improve requirement visibility
- Removed mentioning of 32bit & 64bit. When both are supported, its required to say cause any version they can choose.
- I only 64bit in the case it is a hard requirement. E.g. Windows Phone 8 Emulator

### Checklist


brodybits:
Looks OK from a quick read, approved on my part.

I am actually very hopeful we can drop Windows support in the near future, as discussed elsewhere.

erisu:
@brodybits Thanks

Yes, I feel the same. Trying to not spend a lot of time on this platform and hopefully drop support eventually.

Only doing minimum changes if I notice something, and I though of this after merging in your year old PR #1008. Sorry it took a year to merge it in.

PR #1133, opened 2020-11-06
Added IndexedDB iOS WebView support notes

IgorVNC:
### Platforms affected

Cordova iOS

### Motivation and Context
Cordova IndexedDB documentation is outdated

### Description

I relized that IndedxedDB is supported well on iOS 14, hence need to update Codova docs 

### Testing

I have a project with offline capabilities support. 
Normally we use SQLite, but recently started mooving onto IndexedDB.
I did some simple tests and looks like IndexedDB is supported on iOS14 WebView well

### Checklist

not required

### DIFF
--- a/www/docs/en/dev/cordova/storage/storage.md
+++ b/www/docs/en/dev/cordova/storage/storage.md
@@ -166,6 +166,7 @@ IndexedDB is supported by the underlying WebView on the following Cordova platfo
 
 - Windows (with some limitations)
 - Android (4.4 and above)
+- iOS (14 and above)
 
 ### Windows Limitations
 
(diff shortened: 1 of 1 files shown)


IgorVNC:
Eiger to check under iOS 13 as well, but I do not have iOS device with iOS13

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -166,6 +166,7 @@ IndexedDB is supported by the underlying WebView on the following Cordova platfo
 
 - Windows (with some limitations)
 - Android (4.4 and above)
+- iOS (14 and above)
```
timbru31, line 169: This is simply wrong.

timbru31 (changes requested):
Thanks a lot for your PR, however we already have #1042 which we should aim to merge in instead.

www/docs/en/dev/cordova/storage/storage.md
IgorVNC: ok thanks for the clarification. Then you guys should merge it ASAP cause it produces so much confusion in terms of app architecture decisions

timbru31:
The mentioned PR was just merged 🎉

PR #1172, opened 2021-05-06
doc(android): document `AndroidInsecureFileModeEnabled`

NiklasMerz:
### Platforms affected
Android

### Motivation and Context

Document https://github.com/apache/cordova-android/pull/1222

### Description

Document setting and add warning from https://developer.android.com/reference/android/webkit/WebSettings#setAllowUniversalAccessFromFileURLs(boolean)

### Checklist


jcesarmobile (commented):
I think it should probably say something about backwards compatibility and avoid data loss on existing apps

Also, just noticed there is another section in the file where we put the actual preference tag with an example, can you add it there too?

NiklasMerz:
I am not sure if this description of all config.xml preferences is the right place for that. 

Maybe [here](https://cordova.apache.org/docs/en/10.x/guide/platforms/android/index.html) on the general Android page? Of course we need it in the announce post.

jcesarmobile:
I've fixed the merge conflicts after merging my PR that removed the AndroidX preference.
I edited right before you commented, so you probably didn't see, can you add the preference to the examples section too?

If you don't want to go into detail about the data loss in this PR it's ok, but it should still be about **loading** from file instead of **accessing** file urls, because it's not the same thing.
If you load from file url, you obviously need to access file urls, but you might still want to load from https and access file urls (i.e. take a picture from the camera and assign it to a img tag) and can wrongly enable this preference to do so.

BTW, that example made me think, do we have some helper like on iOS to convert file urls like the one you get from camera to https urls? would the assets loader be able to load those urls somehow?

www/docs/en/dev/guide/platforms/android/upgrade.md
```diff
@@ -27,6 +27,30 @@ Most of these instructions apply to projects created with an older set
...
+discouraged).
+
+### Breaking changes
+
+Version 10.0.0 introduces a signinificant change how URLs are loaded within the app.
+Prior versions load the apps web files like `index.html` via the file protocol.
+Which means the app starts with the URL `file:///android_asset/www/index.html`. Loading `file:///` URLs is considered insecure
+and [Android has deprecated support](https://developer.android.com/reference/android/webkit/WebSettings#setAllowUniversalAccessFromFileURLs(boolean)).
+Cordova Android 10.0.0 now uses an Android API called `WebViewAssetLoader` to load web content via the HTTP(S) scheme (`https://localhost`) by default.
+Therefore the app now starts with the URL `https://localhost/` instead of `file:///android_asset/www/index.html`. Because this is a new origin you might encouter data loss and you need to migrate your web data (Localstorage, IndexedDB etc).
```
NiklasMerz, line 51: We could add more information around migration and data loss once we have a good suggestion?

NiklasMerz:
> I've fixed the merge conflicts after merging my PR that removed the AndroidX preference.
> I edited right before you commented, so you probably didn't see, can you add the preference to the examples section too?

Done

> If you don't want to go into detail about the data loss in this PR it's ok, but it should still be about **loading** from file instead of **accessing** file urls, because it's not the same thing.

I changed the wording and found the right page to describe that. Maybe others can jump in and help to improve it.
 
> If you load from file url, you obviously need to access file urls, but you might still want to load from https and access file urls (i.e. take a picture from the camera and assign it to a img tag) and can wrongly enable this preference to do so.
> 
> BTW, that example made me think, do we have some helper like on iOS to convert file urls like the one you get from camera to https urls? would the assets loader be able to load those urls somehow?

TBH I don't really what happens with camera and other files. Maybe it still works? I need to test it out.

jcesarmobile (approved):

erisu (approved):

PR #1194, opened 2021-08-21
docs!: remove translations

erisu:
### Motivation and Context

* remove translated docs
* redirect all non-English translated docs to English if exisiting. Pages that no longer exist goes to a 404


breautek (approved):

purplecabbage (changes requested):
This pr should not include its own announcement.  Lets do that seperately.
Please remove the blog post commits.

The rest looks great!

dblotsky:
@purplecabbage @erisu what is the rollout plan for this effort?

I remember about [20 months ago](https://github.com/apache/cordova-docs/pull/1015#issuecomment-576045882) we had this rough outline:

1. Announcing the deprecation in a blog post,
1. Adding the redirects, deprecation pages, etc.,
1. Actually deleting the translated pages (since the redirects made them unreachable), and
1. Deleting the code that supports translations.

Is the plan now the same?

erisu:
@purplecabbage I removed the blog announcement from this PR and moved it here: https://github.com/apache/cordova-docs/pull/1195

erisu:
@dblotsky 

> 1. Announcing the deprecation in a blog post

#1195

> 2. Adding the redirects, deprecation pages, etc.,

Done in this PR.

Redirect commit: https://github.com/apache/cordova-docs/pull/1194/commits/b6fa75e8b3056c32abe557415b6e2dcf51e1e14e

> 3. Actually deleting the translated pages (since the redirects made them unreachable), and

Done in this PR.

> 4. Deleting the code that supports translations.

Done in this PR.

erisu:
@purplecabbage merge it in when you ready.

dblotsky:
@erisu sorry, I should've clarified: I mean the plan where we do these things with *some time between them*, so there is time to respond to user feedback, any unexpected fallout, etc.

www/.htaccess
```diff
@@ -129,3 +129,6 @@ RewriteCond %{HTTPS} !=on
 # Redirect to https version
 RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
 
+# Redirect all docs translations to their English version
+RewriteCond $1 !=en
+RewriteRule ^.*docs/(\w\w(?:-\w\w)?)/(.*)$ {{site.baseurl}}/docs/en/$2 [R=301,L]
```
dblotsky, line 134: 1. Please consider using a 302 (temporary redirect) first, and then changing it to a 301 (permanent redirect) later. If you get the 301 wrong, it will be almost impossible to fix.
2. We have a mechanism for redirecting one language to another: https://github.com/apache/cordova-docs/blob/082c8d213615a012422e4f435bf6d4bcb846881e/www/.htaccess#L66. Have you considered just using that?

tools/bin/gen_versions.js
```diff
@@ -24,18 +24,7 @@ var util = require('./util');
...
-    fr: 'Français',
-    it: 'Italiano',
-    ja: '日本語',
-    ko: '한국어',
-    pl: 'Polski',
-    ru: 'Русский',
-    sl: 'Slovene',
-    'zh-cn': '简体中文',
-    'zh-tw': '繁體中文'
+    en: 'English'
```
dblotsky, line 27: Wow, is this change all that's required to get the tooling to remove translations?

NiklasMerz:
I would prefer to remove the translations as quick and easy as possible. It's just not worth it to spend the huge effort on a long rollout. 

IMHO the are not useful at all and we don't have enough people to work on this.

@erisu Thank you for your effort. You are probably the most active contributor and I appreciate your work to make your and our life easier with stuff like cleaning up docs.

dblotsky:
@NiklasMerz I totally agree. I absolutely support limiting the effort that goes into this.

I'm recommending that we introduce a **delay**, not any extra effort. I propose that it would be best to give Cordova's users (or at least the readers of the blog) a chance to voice concerns and feedback on a large change like this.

Is there something driving the urgency (like, has to happen now instead of 2 weeks from now) of this change?

NiklasMerz:
Pushing the blog post out and deleting the translations two weeks later
sounds like a good idea. I support this. 

I remembered a rollout plan with multiple steps and long wait times that
was to complex in my opinion.

purplecabbage (approved):

PR #1197, opened 2021-08-22
doc(archive-20210820): add snapshot

erisu:
### Motivation and Context

Archive language docs.

This PR will push the archive docs directly to the deployed website.

This PR should not be merged in before https://github.com/apache/cordova-docs/pull/1196. We need to sure that the archive directory is not removed by the deploy process. This is handeled in the https://github.com/apache/cordova-docs/pull/1196 PR.


dblotsky:
@erisu is this part of the effort to obsolete the translations?

1. If so, does that mean that the archive will be "pre-rendered", and will no longer get future template/styling changes?
2. Also, will it still reference the same CSS files as the main site, or will those also get archived?

IMO either is fine, but just want to confirm that this is indeed what will be happening.

erisu:
> is this part of the effort to obsolete the translations?

I personally dont care which way this PR goes. It is just an alternative option for keeping our gibberish Lorem Ipsum equivalent translated docs.

> If so, does that mean that the archive will be "pre-rendered", and will no longer get future template/styling changes?

It is an extract of the already generated content from "cordova.apache.org".

https://erisu.github.io/cordova-docs-archived/

> Also, will it still reference the same CSS files as the main site, or will those also get archived?

No the assets are all archived. That includes CSS, Images, Favicon, JS.... everything.

erisu:
IF we merge in this PR, then that means the blog post would be updated. IT would say, "we have deprcated and archived our language docs". We would not say "removed".

With that in mind, I think it technically means we do not need to wait 2 weeks to hear users voice or concerns. We are not removing.

It would also mean the htaccess rewrite rule just need to redirect non-English docs to `/archive/docs/<lang>/...` instead.

dblotsky:
@erisu @purplecabbage to me personally, the idea of archiving is *extremely* appealing. What do you think of switching to that route?

@erisu I would still advocate for at least a few days' delay to give readers time to at least feel heard. I think if we do the change at the same time as the announcement, some folks will be upset.

breautek (approved):

Issue #1200, opened 2021-08-26
Website branding compliance


breautek:
From a recent PMC email notice:

>IMPORTANT: PMCs who list "powered by" companies on their Project sites (organizations that are using your project --in their products/solutions or otherwise) MUST NOT BE DOING SO ON THE PROJECT HOME PAGE.

>WHY: the ASF is strictly vendor-neutral. Listing company names and/or logos on a project's home page is disallowed as it can imply "ownership" or preferred status or otherwise imbalanced relationship. Whether this is the case or not, it is not acceptable. The policy is the same as "thanking" companies for their support of a project http://apache.org/foundation/marks/linking#projectthanks

>WHAT WE'RE NOT SAYING: please don't misunderstand the "no third party logos on your homepage" policy. We *love* projects promoting their users! Many of you have commented that we always include user organizations in our press releases --one of the first things that members of the media and analyst community ask is, "who uses you?". We're proud of all our projects and your collective efforts: we want to ensure that everyone is being fairly represented and considered.

Cordova is explicitly listed as an offender of this policy. I believe both "Cordova Tools" and "Cordova App Showcase" are the problematic sections.

As outlined in the email, having these sections is okay, but they must not be on the homepage.

So the question is do we keep it, and if so, where do we move it? Sally does encourage having this "who uses us" list, so i think it is a good idea to keep it, we just need to decide where is the best place to move it.

PR #1201, opened 2021-08-29
Third party tooling

breautek:
Closes https://github.com/apache/cordova-docs/issues/1200

Completely removes app show case.
Moves the Cordova tooling to a documentation page labelled "Third-party Tools".

The page contains a small spiel about how the list and content is maintained externally by the community.


NiklasMerz (approved):
Dropping the app showcase is a good idea

breautek:
Will be merging this in with the next couple of days or so if there are no objections.

PR #1222, opened 2022-01-01
Fix CocoaPods section grammar

266-750Balloons:
Previously, it said that 'The CocoaPods tools _is_ needed", which is grammatically incorrect.

In order for it to be correct, it needs to say 'The CocoaPods tools _are_ needed'.


breautek:
Thank you. The changes should appear on our dev version of the docs usually within an hour of the merge.

This change will automatically be included on our next docs release.

PR #1245, opened 2022-05-06
small fix category in doap file for better https://projects.apache.org/projects.html?category rendering

ebarboni:
The following page https://projects.apache.org/projects.html?category expose cordova own category instead of belonging to existing mobile and library categories
This is caused by wrong link in category element in the doap file.

### DIFF
--- a/www/infra/doap_Cordova.rdf
+++ b/www/infra/doap_Cordova.rdf
@@ -39,8 +39,8 @@
     <programming-language>C++</programming-language>
     <programming-language>C#</programming-language>
     <programming-language>node.js</programming-language>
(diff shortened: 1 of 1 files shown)


erisu:
@ebarboni I am not 100% positive on these changes as it feels incorrect.

The URLs that you changed to does not exist.

Going to the URLs displays the following error message:
```
Not Found
The requested URL was not found on this server.
````

But on https://projects.apache.org/projects.html?category, I do see the new categories that your referring to that is incorrect.

Even if the URLs displays `Not Found`, is it still accurate?

dpogue:
> Even if the URLs displays `Not Found`, is it still accurate for DOAP?

It seems to match what other projects are doing, and they're showing up in the right categories as intended.

PR #1246, opened 2022-05-09
Correct "plugins" section syntaxe in package.json

jofravelo:
### Platforms affected
Android, iOS

### Motivation and Context
Having hasd to use package.json to declare plugins in my project, I noticed that the array form was wrongly communicating the names of the plugins as numbers, so noticed that cordova is specting it to be an object/map and not an array.

### Description
Only changed the list of plugins from an array to an object.

### Testing
- Execute the prepare step as usual declaring plugins, only in package.json, as an array. It should fail with a message like the following:
  -   `Failed to restore plugin "0". You might need to try adding it again. Error: CordovaError: Cannot find plugin.xml for plugin "0". Please try adding it again.
Discovered plugin "2". Adding it to the project`
- Then, change it to an object. It should work now.

### Checklist

- [X ] I've updated the documentation if necessary

### DIFF
--- a/www/docs/en/dev/platform_plugin_versioning_ref/index.md
+++ b/www/docs/en/dev/platform_plugin_versioning_ref/index.md
@@ -199,9 +199,9 @@ After running the above command, the **`package.json`** should contain something
 
 ```json
 "cordova": {
-  "plugins": [
-    "cordova-plugin-device"
-  ]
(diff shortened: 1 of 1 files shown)


breautek (approved):

breautek:
Thank you for contribution.

PR #1257, opened 2022-07-27
Add Fig as an autocomplete method

brendanfalk:
[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including Cordova. It looks like this:

image

We think this will help users become more efficient with Cordova CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!

### DIFF
--- a/www/docs/en/dev/guide/cli/index.md
+++ b/www/docs/en/dev/guide/cli/index.md
@@ -70,6 +70,16 @@ To install the `cordova` command-line tool, follow these steps:
    `cordova` on the command line with no arguments and it should
    print help text.
 
+### Autocomplete
+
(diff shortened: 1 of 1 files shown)


erisu:
@brendanfalk nice tool!

IMO, I think you should add your content about this tool to the bottom of the `www/docs/en/dev/third-party/index.md` file. This file is where all the third-party tools and libraries are listed.

The file I mentioned above is this page specifically: https://cordova.apache.org/docs/en/dev/third-party/index.html

The file that you added it to is specifically about creating a Cordova app. It focuses on the minimum requirements for installing and using CLI commands.

I don't think your tool would not fall under the minimum requirement.

erisu:
Closed as unreponsive

PR #1258, opened 2022-08-31
docs(splash-screen): update Android 12 bug note

jcesarmobile:
### Platforms affected
Android

### Description

The note is incorrect, it's an Android 12 bug affecting real devices too, not just the emulator.

### Checklist


breautek (approved):
Good to know.

The original PR was only ever tested against simulators, so it wasn't known at the time if the bug actually affected physical devices.

jcesarmobile:
removed display from `will display show` as I think it's a typo, looks like it was going to be `will display` or `will show`, but not both. Used show as it's what's used on the title of the note.

Issue #1232, opened 2022-02-11
Some change requests for the "config.xml" page


DAnn2012:
Some change requests for the "config.xml" page:

![content](https://user-images.githubusercontent.com/1197819/153643703-9092352f-828a-4250-9922-08281c698d97.png)

**content**
The description is duplicated and the attributes are not shown in a table.

<hr>

![access](https://user-images.githubusercontent.com/1197819/153644053-dbb3e40c-6642-417a-933e-462b3f947e82.png)

**access**
The description is duplicated and the attributes are not shown in a table. The "Whitelist Guide" link is not working.

<hr>

![allow-navigation](https://user-images.githubusercontent.com/1197819/153644533-b5893dea-3362-4169-829b-384ed46caa20.png)

**allow-navigation**
The description is duplicated and the attributes are not shown in a table. There is a reference to the old cordova-plugin-whitelist

<hr>

![allow-intent](https://user-images.githubusercontent.com/1197819/153644888-07ac5e0c-6da6-49ad-893c-d3c89abda7ad.png)

**allow-intent**
The description is duplicated and the attributes are not shown in a table. There is a reference to the old cordova-plugin-whitelist

<hr>

![edit-config-config-file](https://user-images.githubusercontent.com/1197819/153645050-b62d5551-7754-42e0-a7ad-17b35bf86dba.png)

**edit-config** & **config-file**
The links to edit-config & config-file for plugin.xml is not working.

<hr>

![AndroidInsecureFileModeEnabled](https://user-images.githubusercontent.com/1197819/153645457-6825bd03-2e19-407e-90e9-0d762d48d24e.png)

**AndroidInsecureFileModeEnabled**
file: is indicated once with three slashes and once with two slashes

<hr>

![KeepRunning](https://user-images.githubusercontent.com/1197819/153645765-77974983-a589-4c86-b4b1-0e9d9cb514d4.png)

**KeepRunning**
The links to pause event is not working.

<hr>

![Sample config.xml](https://user-images.githubusercontent.com/1197819/153645921-dfb57b61-8246-4701-ac2b-2bf5ace3cdef.png)

**Sample config.xml**
There is a reference to the old cordova-plugin-whitelist

<hr>

![Min/Max Version](https://user-images.githubusercontent.com/1197819/153646032-a8a2d8a2-27e1-454d-839d-d4e323a9e983.png)

**Min/Max Version**
The text shows the HTML "ul" and "li" tags

Thanks.

[...]

PR #1263, opened 2022-09-27
Removed some unwanted carriage returns in the Config.xml documentation page

DAnn2012:
### Platforms affected
`Config.xml` page on `cordova.apache.org` https://cordova.apache.org/docs/en/11.x/config_ref/index.html

### Motivation and Context
Removed some unwanted carriage returns (which could be the cause of some of the problems of not displaying the table form described here: #1232 ) and the obsolete reference to plugin `cordova-plugin-whitelist` in the example of the `config.xml` file

### Description
Removed some unwanted carriage returns (which could be the cause of some of the problems of not displaying the table form described here: #1232 ) and the obsolete reference to plugin `cordova-plugin-whitelist` in the example of the `config.xml` file


breautek:
Generally speaking we should be updating the `dev` version. Updating the release versions can also be desirable sometimes, but if the issue is in a released version, chances are it's also in the `dev` version as well and the issue will reappear on our next release/snapshot if it's not corrected inside the [dev](https://github.com/apache/cordova-docs/tree/master/www/docs/en/dev) folder.

I feel like most of these changes probably needs to be corrected in the `dev` folder as well, can you confirm?

DAnn2012:
Hello @breautek 

You're right. I made the changes in the `dev` folder as well.

breautek (approved):

breautek:
Thank you providing the much needed attention to the documentation :)

PR #1277, opened 2022-12-16
Clarify default value for packageType build flag

jkronborg:
### Platforms affected

Android

### Motivation and Context

Documentation clarification that will save developers some time.

### Description

Document that the default value for the `packageType` build flag is `bundle` for release builds.

### Testing

No testing done.

### Checklist


breautek (commented):
I think this is an improvement. I noticed that you backported the change to 10.x version of the docs. I think we can also apply the change to the 11.x version of the docs as well since it would apply there too.

The `dev` version will cover the next time we do a snapshot for the new version.

jkronborg:
@breautek 

> I think this is an improvement. I noticed that you backported the change to 10.x version of the docs. I think we can also apply the change to the 11.x version of the docs as well since it would apply there too.
> 
> The `dev` version will cover the next time we do a snapshot for the new version.

I meant to do that but somehow missed it. Done now. Thanks!

breautek (approved):
Awesome! :D

breautek:
The changes will appear live usually within an hour.

PR #1285, opened 2023-02-11
docs(android): support for themed icons

liyamahendra:
### Platforms affected
- android

### Motivation and Context

Android 13 added support for Themed Icons. Since themed icons aren't yet supported on Cordova Android, adding support for Themed icons.

closes https://github.com/apache/cordova-android/issues/1450

### Description

Added details about the `monochrome` attribute for Adaptive icons

### Testing

n/a

### Checklist

### DIFF
--- a/www/docs/en/dev/config_ref/images.md
+++ b/www/docs/en/dev/config_ref/images.md
@@ -54,7 +54,9 @@ different screen resolutions.
 
 ## Android
 
-On Android, instead of using a single image for an icon, you can use two images (background and foreground) to create an **Adaptive Icon**. WIth the support for themed icons from Android 13, you can use three images (background, foreground and monochrome) to create an **Adaptive Icon**. To use Adaptive Icons, you need to have installed at least version 9.0.0 of **Cordova** and version 8.0.0 of **Cordova-Android**.
+Instead of using a single image for an icon, you can use two images (background and foreground) to create an **Adaptive Icon**. To use Adaptive Icons in Cordova, at minimum **Cordova CLI** 9.0.0 and **Cordova-Android** 8.0.0 is required.
+
+Android 13 has introduced themed icons which are monochrome images appended to the existing **Adaptive Icons**. To use Themed Icon in Cordova, at minimum **Cordova CLI** 12.0.0 and **Cordova-Android** 12.0.0 is required.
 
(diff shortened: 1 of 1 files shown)


liyamahendra:
@breautek just tagging you here. Kindly tag anyone that need to verify the changes please.

www/docs/en/dev/config_ref/images.md
```diff
@@ -54,27 +54,28 @@ different screen resolutions.
 
 ## Android
 
-On Android, instead of using a single image for an icon, you can use two images (background and foreground) to create an **Adaptive Icon**. To use Adaptive Icons, you need to have installed at least version 9.0.0 of **Cordova** and version 8.0.0 of **Cordova-Android**.
+On Android, instead of using a single image for an icon, you can use two images (background and foreground) to create an **Adaptive Icon**. WIth the support for themed icons from Android 13, you can use three images (background, foreground and monochrome) to create an **Adaptive Icon**. To use Adaptive Icons, you need to have installed at least version 9.0.0 of **Cordova** and version 8.0.0 of **Cordova-Android**.
```
erisu, line 57: I think it might be better to split this change into a separate paragraph. Stuck in the middle sounds confusing because the following line talks about the minimum requirement for Adaptive Icons in Cordova, while the requirements for monochrome are different.

Here is maybe an example, but the wording should be confirmed.

```suggestion
Instead of using a single image for an icon, you can use two images (background and foreground) to create an **Adaptive Icon**. To use Adaptive Icons in Cordova, at minimum **Cordova CLI** 9.0.0 and **Cordova-Android** 8.0.0 is required.

Android 13 has introduced themed icons which are monochrome images appended to the existing **Adaptive Icons**. To use Themed Icon in Cordova, at minimum **Cordova CLI** 12.0.0 and **Cordova-Android** 12.0.0 is required.
```

liyamahendra: Using a separate paragraph for Themed icons is way better, thank you! And thank you for adding the `v12.0.0` requirement.

I think @erisu your suggestion is apt; unless there is anything else that you think I should consider, I think this is good to go. Kindly let me know.

breautek (approved):

erisu (approved):
LGTM

PR #1291, opened 2023-03-22
docs(ios): document new InspectableWebview setting

BBosman:
### Platforms affected

iOS

### Motivation and Context

Adds documentation for the changes introduced in https://github.com/apache/cordova-ios/pull/1300 and https://github.com/apache/cordova-plugin-inappbrowser/pull/995.

### Description

Adds documentation for the `InspectableWebview` preference setting.

### Testing

English isn't my native language, so please feel free to suggest better wording.

### Checklist


dpogue (approved):

www/docs/en/dev/config_ref/index.md
```diff
@@ -311,6 +311,7 @@ GradlePluginKotlinEnabled (boolean) <br/> ==Android== | *Default: false* <br/>
 GradlePluginKotlinCodeStyle (string) <br/> ==Android== | *Default: official* <br/> Allowed values: official, obsolete<br/>  Sets how the Kotlin code is formatting for readability.
 GradlePluginKotlinVersion (semver) <br/> ==Android== | *Default: 1.3.50* <br/> Set the version of the Kotlin Gradle plugin to be used.
 InAppBrowserStorageEnabled (boolean) <br/> ==Android== | *Default: true* <br/>  Controls whether pages opened within an InAppBrowser can access the same localStorage and WebSQL storage as pages opened with the default browser.
+InspectableWebview(boolean) <br/> ==iOS== | *Default: depends on Debug/Release build* <br/>  As of iOS 16.4 the webview is no longer inspectable by default. Cordova now allows inspection on Debug builds, but not on Release builds by default. Set this to true to force inspectability regardless of build settings.
```
breautek, line 314: I feel like it's not clear enough that the setting only is honoured on 16.4 or later.

If I recall correctly, iOS 16.3 and earlier, regardless of iOS SDK used to build will always have the old behaviour (always inspectable when in debug mode, always disabled when n release mode).

Additionally, I feel like the current text is written more in a blog format. We don't need to explain the rationale here, we just need to document how to use the preference flag and potential nuances surrounding it.

```suggestion
InspectableWebview(boolean) <br/> ==iOS== | *Default: depends on Debug/Release build* <br/>  On iOS 16.4 or later, enables or disables the webview inspector. Defaults to `true` on debug builds and `false` on release builds. This preference is ignored on iOS 16.3 and earlier; the inspector is always enabled on debug builds and always disabled on release builds.
```

Don't feel like you need to accept this suggestion, I'm open to criticism or feedback.

Ref: https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/

breautek: Additionally, I just created a sister PR for the android (https://github.com/apache/cordova-android/pull/1589), so I think we can add the `==Android==` tag, if you don't mind

PR #1325, opened 2023-07-19
doc: add preference scheme & hostname

erisu:
### Motivation and Context

Missing docs for `preference` `scheme` and `hostname`.

### Description

Added doc to the config.xml page for `preference` `scheme` and `hostname`.

### Testing

`npm run serve`


breautek (approved):

dpogue (approved):
Only suggestion would be to expand on what "non-reserved scheme" means for iOS, but I don't think there's a list of them in any Apple documentation. They just say not to use any scheme that WebKit already supports.

dpogue (approved):

PR #1359, opened 2024-09-11
Fix "spcified" to "specified" in index.md

hugo-levet:


NiklasMerz:
Thank you for finding this. I think this file gets pulled from the cordova-cli repo so this typo should be fixed here: https://github.com/apache/cordova-cli/blob/b57fb4543b7a9c8a912a922f3a34d88dc5fb65a2/doc/readme.md?plain=1#L536

PR #1360, opened 2024-10-10
doc(android): document ZSH setup for macOS Catalina and newer

GitToTheHub:
On macOS Catalina and newer, Apple switched from Bash to ZSH. The new file for ".bash_profile" is ".zprofile".

Also the AnroidSDK path is different on a current macOS with the latest Android Studio, but I don't know since when this was changed, or if it was always like this.

### Platforms affected
Android on macOS

### Motivation and Context
Environment variables on newer macOS system must stored on `.zprofile` instead of `.bash_profile`

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/android/index.md
+++ b/www/docs/en/dev/guide/platforms/android/index.md
@@ -282,29 +282,48 @@ It is also recommended to update the `PATH` environment variable to include the 
 
 _**Note:** The directories above are generally located in the Android SDK ROOT._
 
-#### macOS and Linux
+#### Linux and macOS
 
-On a Mac or Linux, with a text editor, create or modify the `~/.bash_profile` file.
+On Linux or macOS versions prior to Catalina, use any text editor to create or modify the `~/.bash_profile` file.
+<br>On macOS Catalina and newer, create or modify the `~/.zprofile` file, as the default shell has changed.
 
-To set an environment variable, add a line that uses `export` like so (substitute the path with your local installation):
+Add the following line to your shell's profile to set up the `ANDROID_HOME` environment variable.
+    
+**Linux:**
 
 ```bash
-export ANDROID_HOME=/Development/android-sdk/
+export ANDROID_HOME=~/Android/Sdk
 ```
 
-To update your `PATH`, add a line resembling the following (substitute the paths with your local Android SDK installation's location):
+**macOS:**
+
+```bash
+export ANDROID_HOME=~/Library/Android/sdk
+```
+
+The above examples point to the standard paths where the Android SDK is typically stored. Depending on your environment configuration, the path may differ. Be sure to confirm that the path exist or adjust them to match your local installation.
+
+After setting the `ANDROID_HOME` environment variable, update the `PATH` variable to include directories containing various Android-related binaries.
 
 ```bash
 export PATH=$PATH:$ANDROID_HOME/platform-tools/
 export PATH=$PATH:$ANDROID_HOME/cmdline-tools/latest/bin/
-export PATH=$PATH:$ANDROID_HOME/build-tools
+export PATH=$PATH:$ANDROID_HOME/build-tools/
 export PATH=$PATH:$ANDROID_HOME/emulator/
 ```
 
 Reload your terminal to see this change reflected or run the following command:
 
+**Linux and macOS older then Catalina:**
+
 ```bash
 source ~/.bash_profile
+```
+
+**macOS Catalina and newer:**
+
+```bash
+source ~/.zprofile
 ```
 
 #### Windows


www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -284,14 +284,21 @@ _**Note:** The directories above are generally located in the Android SDK ROOT._
 
 #### macOS and Linux
 
-On a Mac or Linux, with a text editor, create or modify the `~/.bash_profile` file.
+On a Mac older then Catalina or on Linux, create or modify the `~/.bash_profile` file for e.g. with a text editor.
```
erisu, line 287:
```suggestion
On macOS versions prior to Catalina or on Linux, use any text editor to create or modify the `~/.bash_profile` file.
```

www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -284,14 +284,21 @@ _**Note:** The directories above are generally located in the Android SDK ROOT._
...
 ```bash
 export ANDROID_HOME=/Development/android-sdk/
 ```
 
+on macOS
+
+```zsh
+export ANDROID_HOME=~/Library/Android/sdk/
+```
+
```
erisu, line 301: This section may be unnecessary, as the previous line instructs to "substitute the path with your local installation."

The example path, `/Development/android-sdk/`, as far as I am aware is not a standard or default path, even on Linux. It was only for an example. I believe it was intended to account for users who set up their Android development environment with `cmdline-tools` rather than Android Studio.

www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -284,14 +284,21 @@ _**Note:** The directories above are generally located in the Android SDK ROOT._
 
 #### macOS and Linux
 
-On a Mac or Linux, with a text editor, create or modify the `~/.bash_profile` file.
+On a Mac older then Catalina or on Linux, create or modify the `~/.bash_profile` file for e.g. with a text editor.
+On macOS Catalina and newer, create or modify the `~/.zprofile`, because it's using ZSH instead of Bash.
```
erisu, line 288:
```suggestion
On macOS Catalina and newer, create or modify the `~/.zprofile` file instead, as the default shell has changed.
```

erisu (changes requested):
I made some minor changes.

Do you have any thoughts on my suggestion about possibly removing part of the added content?

www/docs/en/dev/guide/platforms/android/index.md
dpogue: Maybe something like this is clearer that it's an example placeholder?

```bash
export ANDROID_HOME=/path/to/android-sdk/
```

GitToTheHub: Could there be not suggestions for the different OSs? It is more helpful, if the Guide already takes you to the right place of your SDK.
I overhead the part "substitute the path with your local installation". But the example `ANDROID_HOME=/path/to/android-sdk/`would make it more clear

www/docs/en/dev/guide/platforms/android/index.md
```diff
@@ -284,14 +284,21 @@ _**Note:** The directories above are generally located in the Android SDK ROOT._
...
 #### macOS and Linux
 
-On a Mac or Linux, with a text editor, create or modify the `~/.bash_profile` file.
+On macOS versions prior to Catalina or on Linux, use any text editor to create or modify the `~/.bash_profile` file.
+On macOS Catalina and newer, create or modify the `~/.zprofile` file instead, as the default shell has changed.
 
 To set an environment variable, add a line that uses `export` like so (substitute the path with your local installation):
 
 ```bash
 export ANDROID_HOME=/Development/android-sdk/
```
GitToTheHub, line 293:
```suggestion
export ANDROID_HOME=/path/to/android-sdk/
```

www/docs/en/dev/guide/platforms/android/index.md
erisu: If we wanted to suggest for different OS, we could update the entire section like this?

```
    Add the following line to your shell's profile to set up the `ANDROID_HOME` environment variable.
    
    **macOS:**
    
    ```bash
    export ANDROID_HOME=~/Library/Android/sdk
    ```
    
    **Linux:**
    
    ```bash
    export ANDROID_HOME=~/Android/Sdk
    ```
    
    The above examples point to the standard paths where the Android SDK is typically stored. Depending on your environment configuration, the path may differ. Be sure to confirm that the path exist or adjust them to match your local installation.
    
    After setting the `ANDROID_HOME` environment variable, update the `PATH` variable to include directories containing various Android-related binaries.
    
    ```bash
    export PATH=$PATH:$ANDROID_HOME/platform-tools/
    export PATH=$PATH:$ANDROID_HOME/cmdline-tools/latest/bin/
    export PATH=$PATH:$ANDROID_HOME/build-tools/
    export PATH=$PATH:$ANDROID_HOME/emulator/
    ```
```

The above:

* Provide example `export` lines for macOS and Linux, pointing to standard paths.
* Mention that these are **standard paths**, but that the actual path **may differ** depending on your environment configuration.
* Remind you to **confirm that each path exists** and to **adjust paths as needed** to match your installation.

I also revised the instructions above the `export PATH` section to improve flow and make it easier to follow. While it’s helpful to provide cut-and-paste examples, it’s equally important that users review each step, as environments can vary.

If you use the above, make sure to fix the indentation from what I posted it. I had to indent the content to prevent GH from rendering the markdown.

GitToTheHub: Ah sorry, I didn't saw Windows has it's own section :)

GitToTheHub (commented):
I took over the suggestions by @erisu and named always first Linux and then macOS. So it's consistent, that always Linux is first described and then macOS.

erisu (approved):

PR #1372, opened 2024-10-23
announcement: cordova-plugin-file@8.1.1

breautek:


www/_posts/2024-10-23-cordova-plugin-file-8.1.1.md
```diff
@@ -0,0 +1,42 @@
...
+cordova plugin add cordova-plugin-file@8.1.1
+```
+
+## Release Highlights
+
+For Android, we corrected an issue that prevented `.toURL()` to be used on `content://` file entries. Now `.toURL()` will return a DOM-usable url on both `file://` and `content://` file entries.
+
+This will be important for upcoming releases planned for the Camera plugin.
+
+Note that this only works when `AndroidInsecureFileModeEnabled` is off. If `AndroidInsecureFileModeEnabled` then `.toURL()` will return the underlying URL and modern Android devices will reject `content://` file paths. If your project is still using `AndroidInsecureFileModeEnabled` it is time to strongly consider migrating to schemes
```
dpogue, line 28: This part reads a bit weird IMO: "If `AndroidInsecureFileModeEnabled` then ..."
Although "If `AndroidInsecureFileModeEnabled` is enabled then ..." would also be weird.

Maybe "If the `AndroidInsecureFileModeEnabled` preference is turned on then ..."

dpogue (approved):

www/_posts/2024-10-23-cordova-plugin-file-8.1.1.md
breautek: I did a little bit of rewording, let me know if it looks good to you if you can.

dpogue: LGTM

PR #1378, opened 2024-11-06
feat!: cleanup & refactor dependency usage

erisu:
### Platforms affected

n/a

### Motivation and Context

- Remove unused dependencies
- Reduce dependencies footprint for node internals

### Description

- Remove `optimist` & replaced with `minimist` + custom wrapper.
- Remove `fs-extra` for `node:fs`
- Remove `chalk` for `node:util` `styleText`
- Prefixed node internal packages with `node:*`
- Remove `ncp` for `fs.cp`
- Remove `gulp-less` - No less files in repo
- Remove `baconjs` - Not used anywhere
- Remove `@babel/core` & `@babel/preset-react` - Not used anywhere.
- Remove `preact` & `preact-compat` - Not used anywhere. Last usage was prior to plugin search page redesign.
- Remove `classnames` - Not used anywhere. Last usage was prior to plugin search page redesign.
- Updated `sass` & `gulp-sass` dependencies
- Updated `semver`
- Updated `@cordova/eslint-config@5.1.0` w/ lint fix
- Bump node engine requirement to `v20.18.0`
- Resolve `sass` lint warnings with `darken` and `lighten`
  - https://sass-lang.com/documentation/breaking-changes/color-functions/

### Testing

- `npm t`
- `npm run serve`

### Checklist


www/static/css-src/_home.scss
```diff
@@ -272,7 +272,7 @@ img#logo_top {
 }
 
 .supported_platforms {
-    background-color: darken(white, 85%);
```
dpogue, line 275: Since this is happening on a static colour `white` we should just figure out what the hex value is and use it directly or store it directly in a variable instead of needing a colour function here.

www/static/css-src/_home.scss
```diff
@@ -293,7 +293,7 @@ img#logo_top {
     .card {
             padding:8px;
             .card_inner {
-                background-color: darken(white, 85%);
```
dpogue, line 296: Here too, since this is happening on a static colour `white` we should just figure out what the hex value is and use it directly or store it directly in a variable instead of needing a colour function here.

dpogue (approved):
It might be possible to replace `minimist` with `node:util/parseArgs` for these simple cases, but I think that would require you to handle checking for required options yourself.

PR #1388, opened 2024-12-26
doc(ios): update requirements & support table

erisu:
### Motivation and Context

Update documentation for `cordova-ios@8.0.0`

### Description

Updated requirements & support table

### Testing

n/a

### Checklist

### DIFF
--- a/www/docs/en/dev/guide/platforms/ios/index.md
+++ b/www/docs/en/dev/guide/platforms/ios/index.md
@@ -50,6 +50,18 @@ The table below outlines Cordova-iOS's prerequisites by version. Additionally, X
     </thead>
     <tbody>
         <tr>
+            <td>8.x</td>
+            <td>13.0</td>
+            <td>
+                <ul>
+                    <li>Cocoapods: >=1.16.0</li>
+                    <li>ios-deploy: >=1.12.2</li>
+                    <li>Node: >=20.5.0</li>
+                    <li>Xcode (xcodebuild): >=15.0.0</li>
+                </ul>
+            </td>
+        </tr>
+        <tr>
             <td>7.x</td>
             <td>11.0</td>
             <td>


www/docs/en/dev/guide/platforms/ios/index.md
```diff
@@ -49,6 +49,18 @@ The table below outlines Cordova-iOS's prerequisites by version. Additionally, X
...
     </thead>
     <tbody>
+        <tr>
+            <td>8.x</td>
+            <td>13.0</td>
+            <td>
+                <ul>
+                    <li>Cocoapods: >=1.16.0</li>
+                    <li>ios-deploy: >=1.12.2</li>
+                    <li>Node: >=16.13.0</li>
```
dpogue, line 59: I think our intention is to target NodeJS 18? But that would put us out of sync with the rest of the Cordova tooling... but I also don't want to delay the iOS release until we're ready to depend on 20 for everything 😕

erisu: I think we can keep it at `>=16.13.0` for now...

Maybe in April/May we could bump the packages to `>=20`?

dpogue: In terms of testing, I think we already upgraded to packages that only support NodeJS 18+ in cordova-ios and we're not testing 16.x in CI anymore, so I don't want to promise 16 support if we aren't actually making sure that it works.

I'd rather go with 18 as a minimum, or make a decision to use 20 (and then avoid another major bump when we bump the tooling?)

erisu: Then lets go with 20.x... I would suggest to create the vote on mailing list since we will be deprecating 18 while its still active in Node.

erisu: Updated to `>=20.5.0` to match with recent PR changes.

dpogue (approved):

PR #1399, opened 2025-02-12
doc(ios): install cocoapods with homebrew

GitToTheHub:
- cocoapods are not installable with the current system ruby version on macOS. When trying to install with `sudo gem install cocoapods`, the error message will appear: `securerandom requires Ruby version >= 3.1.0. The current ruby version is 2.6.10.210`.
- On stackoverflow it is suggested to install it instead with homebrew: https://stackoverflow.com/questions/77339560/error-installing-cocoapodsdrb-requires-ruby-version-2-7-0-the-current-ruby

### Platforms affected
macOS

### Motivation and Context
I tried to setup my machine and got an error when installing cocoa pods

### Checklist

### DIFF
--- a/www/docs/en/12.x-2025.01/guide/platforms/ios/index.md
+++ b/www/docs/en/12.x-2025.01/guide/platforms/ios/index.md
@@ -124,7 +124,7 @@ The [CocoaPods](https://cocoapods.org/#install) tools are needed to build iOS ap
 To install CocoaPods, run the following from command-line terminal:
 
 ```bash
-$ sudo gem install cocoapods
+$ brew install cocoapods
 ```
 
 ## Project Configuration
--- a/www/docs/en/12.x/guide/platforms/ios/index.md
+++ b/www/docs/en/12.x/guide/platforms/ios/index.md
@@ -124,7 +124,7 @@ The [CocoaPods](https://cocoapods.org/#install) tools are needed to build iOS ap
 To install CocoaPods, run the following from command-line terminal:
 
 ```bash
-$ sudo gem install cocoapods
+$ brew install cocoapods
 ```
 
 ## Project Configuration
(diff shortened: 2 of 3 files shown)


erisu:
Do we know if CocoaPods manages the Homebrew package, or is it handled by a third party?

The current guide follows the steps provided on the official CocoaPods website.

In fact, CocoaPods does not recommend using the default macOS system Ruby. Instead, it suggests using a Ruby version manager to install a newer version. For example, the `rvm` package, which can be installed via Homebrew.

In my case, I have a newer version of Ruby installed on macOS:

```bash
% ruby -v
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin23]
```

Here is their official installation guide, which recommends using a Ruby version manager:
[https://guides.cocoapods.org/using/getting-started.html#installation](https://guides.cocoapods.org/using/getting-started.html#installation)

erisu (approved):
I think it should be OK to update the guide to use `brew install cocoapods`.

It seems,

- When running `brew install cocoapods`, Homebrew uses a [portable version of Ruby](https://github.com/Homebrew/homebrew-portable-ruby).
  - Requirement for Ruby comes from the [`depends_on "ruby"`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/c/cocoapods.rb#L19) line in the formula.
- It downloads the CocoaPods [tarball](https://github.com/Homebrew/homebrew-core/blob/master/Formula/c/cocoapods.rb#L4) from the CocoaPods repository.
- This tarball contains the `cocoapods.gemspec` file, which is then [built and installed](https://github.com/Homebrew/homebrew-core/blob/master/Formula/c/cocoapods.rb#L24-L25).

I'm not 100% sure, but if a user had installed Ruby via `brew`, would Homebrew use that version instead of the portable version?

At the very least, it seems to ignore the older system's default Ruby, as your PR description suggests its resolved the `securerandom requires Ruby version >= 3.1.0. The current ruby version is 2.6.10.210` issue.

I'll wait a bit to see if anyone else has questions or comments. But other than that, I think it's OK to merge.

GitToTheHub:
Cool, thanks for the explanation. I learned a little bit :)

NiklasMerz (approved):
I'm fine with that. Most Mac users should have brew and using it is much more reliable.

PR #1408, opened 2025-03-13
fix(RewriteRule): do not redirect for archive or English docs

erisu:
### Motivation and Context

Fix redirect for archived docs

### Description

Updated the `RewriteRules` to follow these steps, in order:

1. Stop rewrite when URL starts with `/archive/`.
2. Stop rewrite when URL starts with  `/docs/en/`.
3. Forward non-English docs to English docs.

Additional Notes:

- Step three has always been designed to redirect old language documentation to the English version.
  - However, it is possible to change this behavior to redirect them to the archived language documentation instead, **if preferred**. (Would be in a separate PR)
  - However, since this redirection behavior has been in place for several years, changing it now may no longer be necessary.
- If old versions of the English documentation are to be archived, the `RewriteRule` should be placed above step two. In this case, it would redirect to archive.

### Testing

Ran a httpd server to test the RewriteRules.

### Checklist


breautek (approved):
lgtm

PR #1416, opened 2025-04-05
Plugin Guide: Notes for --link option

GitToTheHub:
Added notice, that the linking only works, when the plugin is added after a platform was added and that the linking will be lost if a platform is re-added

### Checklist

### DIFF
--- a/www/docs/en/12.x-2025.01/guide/hybrid/plugins/index.md
+++ b/www/docs/en/12.x-2025.01/guide/hybrid/plugins/index.md
@@ -194,7 +194,7 @@ cordova plugin add ../path/to/my/plugin/relative/to/project --link
 ```
 
 This creates a symbolic link instead of copying the plugin files, which enables you
-to work on your plugin and then simply rebuild the app to use your changes.
+to work on your plugin and then simply rebuild the app to use your changes. You have to add the plugin after you added the platform, otherwise the link will not work. The link will also be lost, if you re-add a platform. If you do this you have to re-add the plugin.
 
 ## Validating a Plugin using Plugman
 


www/docs/en/12.x-2025.01/guide/hybrid/plugins/index.md
```diff
@@ -194,7 +194,7 @@ cordova plugin add ../path/to/my/plugin/relative/to/project --link
 ```
 
 This creates a symbolic link instead of copying the plugin files, which enables you
-to work on your plugin and then simply rebuild the app to use your changes.
+to work on your plugin and then simply rebuild the app to use your changes. You have to add the plugin after you added the platform, otherwise the link will not work. The link will also be lost, if you re-add a platform. If you do this you have to re-add the plugin.
```
erisu, line 197:
```suggestion
to work on your plugin and then simply rebuild the app to use your changes. The plugin should be added after the platform is added; otherwise, the link will not work. The link will also be lost if you re-add the platform or restore the project. In that case, you'll need to re-add the plugin to restore the link.
```

How does the above sound?

erisu (changes requested):
I would recommend adding this to the `dev` directory as well to ensure that it will exist in future snapshots

www/docs/en/12.x-2025.01/guide/hybrid/plugins/index.md
erisu: I used a ";" because they both could be stand-alone sentences.

Restoring a project for example is running the command `cordova prepare` after cloning your project's repo and running `npm i`. Or another example is deleting the existing platfroms and plugins directory and running the `cordova prepare` command.

Issue #1210, opened 2021-10-08
Cache storage documentation


mfulton26:
# Feature Request

Details on advantages/disadvantages, supported platforms, etc. for the Cache API supported by browsers, etc.

## Motivation Behind Feature

Cordova has some great docs for helping when making decisions on where and how to store data but doesn't currently provide information saved about the Cache API which can be used in some cases instead of IndexedDB. Documenting this will help folks make informed decisions.

## Feature Description

Document Cache API similar to how localStorage, IndexedDb, Web SQL, etc. are documented.

www/docs/en/dev/cordova/storage/storage.md

## Alternatives or Workarounds

Various sites out there have some information about total size limits on iOS, etc. etc. but it is hard to find what applies to apps running inside Cordova/WKWebView.

breautek:
Based on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Cache), it looks like the Cache API is a service worker feature. I'm pretty sure service workers are **not** supported by the underlying webviews. They are supported in the browser application themselves (e.g. Safari and Chrome) but not the WKWebView / Android System Webview counterpart in which the native apps have access to.

mfulton26:
Yes, if it could be tested and documented that would be very helpful for reference. Especially for iOS.

Shiva903-hue:
Hi, I’d like to work on this issue and contribute documentation around Cache Storage usage in Cordova.

I plan to add a clear, docs-only section explaining:

How Cache Storage is exposed in Cordova WebView (window.caches)

Platform differences and expectations (Android WebView vs iOS WKWebView)

Persistence and eviction caveats

When Cache Storage is appropriate vs when alternatives like IndexedDB should be used

I’ll avoid making assumptions about exact storage limits and will focus on documented behavior and practical guidance.

If this approach looks good, I’m happy to take ownership of the issue and open a PR for it.
Feel free to assign it to me so others don’t duplicate the work.

breautek:
[...]

PR #1455, opened 2026-01-04
docs: document Cache Storage behavior in Cordova WebViews

Shiva903-hue:
This PR adds a Cache Storage subsection to the existing storage documentation.

It summarizes the web-standard Cache Storage API and documents how it behaves
inside Cordova WebViews, with a focus on platform-specific caveats (Android and iOS),
persistence expectations, and guidance on appropriate usage.

This documentation intentionally avoids positioning Cache Storage as a Cordova
feature and instead aligns with existing storage sections.

References #1210

### DIFF
--- a/www/docs/en/dev/cordova/storage/storage.md
+++ b/www/docs/en/dev/cordova/storage/storage.md
@@ -86,65 +86,6 @@ database, and its asynchronous API and search indexes provide performance benefi
 
 IndexedDB is supported by the underlying WebView on all platforms, with known limitations on the `browser` platform.
 
-## Cache Storage
-
-### Overview
-
-The Cache Storage API is a web-standard storage mechanism primarily used for
-storing network request and response pairs. It is commonly used to support
-resource caching and offline-read scenarios, and is often associated with
-Service Worker–based caching strategies.
-
-### Availability in Cordova
-
-Cache Storage is exposed via the standard `window.caches` interface when supported
-by the underlying WebView implementation. Availability and behavior depend entirely
-on the platform WebView, and the API may not be available on all Cordova platforms
-or in all execution contexts.
-
-### Platform behavior and caveats
-
-Behavior of Cache Storage in Cordova applications can differ from typical desktop
-browser environments:
-
-- **Android (Chromium WebView):** Behavior is generally similar to Chrome, but may
-  vary depending on the WebView version shipped with the device.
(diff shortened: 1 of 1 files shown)


www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -187,6 +188,66 @@ For more information, see:
...
+The Cache Storage API is a web-standard storage mechanism primarily used for
+storing network request and response pairs. It is commonly used to support
+resource caching and offline-read scenarios, and is often associated with
+Service Worker–based caching strategies.
+
+### Availability in Cordova
+
+Cache Storage is exposed via the standard `window.caches` interface when supported
+by the underlying WebView implementation. Availability and behavior depend entirely
+on the platform WebView, and the API may not be available on all Cordova platforms
```
GitToTheHub: `and the API may not be available on all Cordova platforms`: This sounds also, like this has something to do exclusively with Cordova but it's not. Maybe it could be `and the API may not be available on all Cordova supported platforms`

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -187,6 +188,66 @@ For more information, see:
...
+
+### Platform behavior and caveats
+
+Behavior of Cache Storage in Cordova applications can differ from typical desktop
+browser environments:
+
+- **Android (Chromium WebView):** Behavior is generally similar to Chrome, but may
+  vary depending on the WebView version shipped with the device.
+- **iOS (WKWebView):** Cache Storage persistence is more restrictive, and cached
+  data may be evicted more aggressively by the system.
```
GitToTheHub: What does this mean `is more restrictive` and `cached data may be evicted more aggressively by the system.`. An example would be good to explain, what this means.

www/docs/en/dev/cordova/storage/storage.md
```diff
@@ -187,6 +188,66 @@ For more information, see:
...
+
+- Storing user-generated or business-critical data
+- Data that requires guaranteed persistence across app restarts or updates
+
+For structured or durable data storage, consider using IndexedDB instead.
+
+
+### References
+
+- [MDN: Cache Storage API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage)
```
GitToTheHub: I would write
- [MDN: CacheStorage API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) - Handles all caches

PR #1463, opened 2026-01-16
fix: normalize URL comparison in TOC highlighting

dheeraj12347:
## Problem
Table of Contents (TOC) highlighting is not working correctly for some pages.

## Root Cause
The URL comparison logic in `toc_recursive_main.html` and `toc_recursive_dropdown.html` was failing due to trailing slash differences between `include.my_entry` and `entry.url`.

Example:
- include.my_entry: `/docs/api/`
- entry.url: `/docs/api`

These look the same but don't match as strings.

## Solution
Normalize both URLs before comparison by removing trailing slashes using Jekyll's `remove` filter.

Added:
```liquid
{% assign normalized_my_entry = include.my_entry | remove: '/' %}
{% assign normalized_entry_url = entry.url | remove: '/' %}
## Solution
Normalize both URLs before comparison by removing trailing slashes using Jekyll's `remove` filter.

Added:
```liquid
{% assign normalized_my_entry = include.my_entry | remove: '/' %}
{% assign normalized_entry_url = entry.url | remove: '/' %}
Then compare:

text
{% if normalized_my_entry == normalized_entry_url %}
Files Changed
www/_includes/toc_recursive_main.html

www/_includes/toc_recursive_dropdown.html

Why This Works
Removes all / characters from both URLs before comparison

Makes the comparison consistent regardless of trailing slashes

Current page highlighting now works correctly

No JavaScript changes needed

### DIFF
--- a/www/_includes/toc_recursive_dropdown.html
+++ b/www/_includes/toc_recursive_dropdown.html
@@ -1,5 +1,25 @@
 {% for entry in include.entries %}
-{% assign entry_active = include.my_entry | remove: '/' == entry.url | remove: '/' %}
+
+{% assign my_entry_url = include.my_entry %}
+{% assign entry_url = entry.url %}
+
+{% if my_entry_url %}
+    {% if my_entry_url | slice: -1 == '/' %}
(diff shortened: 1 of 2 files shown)


GitToTheHub:
Very good :)

PR #1500, opened 2026-05-20
feat(docs): add latest release versions for Android and iOS platform guides

GitToTheHub:
### Platforms affected

- Android
- iOS

### Motivation and Context

- It good to know, what is the maximum minor/patch version for a major
- Generated-By: GPT-5.3-Codex

### Checklist

### DIFF
--- a/www/docs/en/latest/guide/platforms/android/index.md
+++ b/www/docs/en/latest/guide/platforms/android/index.md
@@ -35,6 +35,7 @@ The supported [Android API Levels](https://developer.android.com/guide/topics/ma
     <thead>
         <tr>
             <th>cordova-android Version</th>
+            <th>Latest Release</th>
             <th>Android API-Levels (Android Version)</th>
             <th>Library &amp; Tooling Version</th>
         </tr>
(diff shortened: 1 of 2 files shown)


dpogue (commented):
Not really a fan of this because it's yet another (third) thing that someone needs to remember to update when making a release, which means it will get missed and have outdated information.

GitToTheHub:
It's true, that on a release this would have to be updated like when cordova-ios 8.1.0 would be released. If cordova.apache.org would support NPM badges by shields.io, which it currently not does, like it can be seen for [cordova-plugin-network-information](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-network-information/index.html) due to CSP erros:

```log
Refused to load https://img.shields.io/npm/v/cordova-plugin-network-information/latest?label=Latest%20Release%20(npm) because it appears in neither the img-src directive nor the default-src directive of the Content Security Policy.
```

this could be automated like already done on the README for the platform e.g. [cordova-ios](https://github.com/apache/cordova-ios/blob/master/README.md).

If a major would be created, cordova-docs has to be anyway updated for it, but the npm badge would have to move to the next major.

erisu:
I agree with @dpogue. I don't think it's necessary to add a column listing the current release version on these pages. We already have a dedicated download page that contains the latest release version information. Duplicating that information here would create another place that needs to be maintained and kept in sync. The purpose of this table was to just outline what are the prerequisites by major release version.

GitToTheHub:
Ok was just an idea. Thanks for the responses!

